PR #15, opened 2019-08-01
Feat/9 metaschema changes

andricDu:
* files -> schemas
* re-org of meta and restriction fields to match spec
* display names


blabadi:
can you try adding display name to the registration test schema fixture to get more coverage

andricDu:
Good call

PR #16, opened 2019-08-07
Basic ego support for JWT

andricDu:
This is a quick solution for the auth until ego-token-middleware is fixed and made avaiable for use with the lastest version of ego. 

### Changes
- Grabs ego public key and caches it
- uses JWT bearer token auth for post/put endpoints
- new error types (unauthorized, forbidden)
- updated swagger with bearer token auth


src/services/egoTokenService.ts
```diff
@@ -0,0 +1,71 @@
...
+import memoize from "memoizee";
+import axios from "axios";
+import jwt from "jsonwebtoken";
+import ms from "ms";
+import { NextFunction, Request, Response, RequestHandler } from "express";
+import { ForbiddenError, UnauthorizedError } from "../utils/errors";
+
+const getKey = memoize(
+    async (egoURL: string) => {
+        const response = await axios.get(urlJoin(egoURL, "oauth/token/public_key"));
```
blabadi: why the other part of the url is hardcoded ?

src/services/egoTokenService.ts
```diff
@@ -0,0 +1,71 @@
...
+        return response.data;
+    },
+    {
+        maxAge: ms("1h"),
```
blabadi: cool shit man

src/services/egoTokenService.ts
```diff
@@ -0,0 +1,71 @@
...
+            const { authorization: authorizationBody }: any = req.body || {};
+
+            const authorization = authorizationHeader || authorizationBody;
+            const bearerToken: string = authorization ? authorization.split(" ")[1] : req.query.key;
+
+            let valid = false;
+            try {
+                valid = !!(bearerToken && (await verifyEgoToken(bearerToken, egoURL)));
+            } catch (e) {
+                valid = false;
```
blabadi: add debug/info logging for the error ? could be useful detecting suspicious behavior

src/config/swagger.json
```diff
@@ -9,6 +9,13 @@
     "http",
```
blabadi: switch this to swagger 3: https://mermade.org.uk/openapi-converter or create the ticket to do that next sprint please

src/services/egoTokenService.ts
andricDu: good point, ill remove this and chop off one dependency.

src/services/egoTokenService.ts
andricDu: :+1:

src/config/swagger.json
andricDu: ill be doing this next sprint but good call.

andricDu:
I need to fix unit tests so that they don't break. Auth/Enabled/Disabled flag would be good.

src/services/egoTokenService.ts
blabadi: you are using 4 libraries in one line and you shamed me for adding yaml2js .. :ok_hand: :joy:

andricDu: :nauseated_face:

src/config/swagger.json
andricDu: https://github.com/overture-stack/lectern/issues/17

blabadi:
> I need to fix unit tests so that they don't break. Auth/Enabled/Disabled flag would be good.

I would add a test to make sure the security logic works as expected, specially if someone fiddles with it later, please add a ticket for that

PR #42, opened 2019-09-17
updated logging, mongo config

andricDu:
Properly configured Winston logging with colourized output

image

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -6,6 +6,7 @@ import { errorHandler } from "./utils/errors";
 import * as swaggerUi from "swagger-ui-express";
 import * as swagger from "./config/swagger.json";
 import ego from "./services/egoTokenService";
+import logger from "./config/logger";
 
 
 /**
@@ -37,6 +38,8 @@ app.use(bodyParser.urlencoded({
 
 swagger["info"]["version"] = process.env.npm_package_version;
 app.use("/api-docs", swaggerUi.serve, swaggerUi.setup(swagger));
+
+logger.info("Swagger setup... done");
 
 app.get("/", (_, res) => {
   const details = {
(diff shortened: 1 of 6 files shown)


src/app.ts
```diff
@@ -38,6 +39,8 @@ app.use(bodyParser.urlencoded({
 swagger["info"]["version"] = process.env.npm_package_version;
 app.use("/api-docs", swaggerUi.serve, swaggerUi.setup(swagger));
 
+logger.info("Swagger setup... done");
```
joneubank: nice to have: console out the URL for swagger

src/server.ts
```diff
@@ -16,12 +47,10 @@ mongoose.connect(constructMongoUri(), { useNewUrlParser: true }).then(
  * Start Express server.
  */
 const server = app.listen(app.get("port"), () => {
-  console.log(
-    "  App is running at http://localhost:%d in %s mode",
-    app.get("port"),
-    app.get("env")
+  logger.info(
+    `App is running at http://localhost:${app.get("port")} in ${app.get("env")} mode`
```
joneubank: 👍

src/utils/errors.ts
```diff
@@ -78,6 +79,10 @@ export const errorHandler = (err: Error, req: Request, res: Response, next: Next
             res.status(400);
             break;
         default:
+            logger.error("Internal Server Error");
+            logger.error(err.name);
+            logger.error(err.message);
+            logger.error(err.stack);
```
joneubank: These four messages could be separated by logs from other requests happening at same time, even another error causing debug confusion... perhaps you want these 4 messages combined into one?

andricDu:
Thanks @joneubank

PR #54, opened 2019-10-23
Feature/43 ranges

blabadi:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:12.5.0-alpine
```
blabadi: i think we should move from node 8 since it's reaching it's LTS end

src/config/MetaSchema.json
```diff
@@ -65,19 +81,195 @@
...
+                        {
+                            "required": [
+                                "min"
+                            ],
+                            "properties": {
+                                "min": {
+                                    "type": "integer"
+                                },
+                                "exclusiveMin": {
+                                    "not": {}
```
blabadi: maybe i don't need the not if I use `additionalProperties : false`, will test that

Dockerfile
andricDu: :+1:

src/config/MetaSchema.json
```diff
@@ -65,19 +81,195 @@
...
+                            ],
+                            "properties": {
+                                "min": {
+                                    "not": {}
+                                },
+                                "exclusiveMin": {
+                                    "not": {}
+                                },
+                                "max": {
+                                    "not": {}
```
andricDu: why the `not`?

PR #57, opened 2019-10-30
Feature/better diff

blabadi:


src/diff/DictionaryDiff.ts
```diff
@@ -11,6 +11,121 @@ interface FieldDiff {
     diff: any;
 }
 
+
```
blabadi: mention source : https://stackoverflow.com/questions/8572826/generic-deep-diff-between-two-objects/31010495

.env.example
```diff
@@ -3,9 +3,12 @@ PORT=3000
 
 # Mongo Configuration
 MONGO_HOST=
+MONGO_PORT=
 MONGO_DB=
 MONGO_USER=
 MONGO_PASS=
+# this takes precedence if provided over the previous entries (easier for local dev)
+MONGO_URL=
```
andricDu: :+1:

PR #62, opened 2019-12-12
🎨 Updating and applying Prettier code format

joneubank:
Made the prettier config match other overture TS projects and our team standard. Applied formatting to every file.

**No code changes here, only formatting**


andricDu:
dev dependiency? conflics between tslint and prettier?

joneubank:
Will add prettier as dev-dependency.

PR #83, opened 2020-04-15
add isArray property to field

wajiha-oicr:
Not actually defining it as an Array because within the tsv it will still just be a string, (albeit with comma separation)

### DIFF
--- a/src/config/MetaSchema.json
+++ b/src/config/MetaSchema.json
@@ -336,9 +336,6 @@
       "properties": {
         "required": {
           "type": "boolean"
-        },
-        "isArray": {
-          "type": "boolean"
         }
       },
       "additionalProperties": false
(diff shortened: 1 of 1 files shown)


wajiha-oicr:
> The MetaSchema has restrictions broken down, so this new property should be included in `booleanRestrictions.properties` .

done in https://github.com/overture-stack/lectern/pull/83/commits/0eecaab8b4053f157eae9b0a0fce46ed2b07bcab

tested and dictionary is accepting fields with or without the isArray property

src/config/MetaSchema.json
```diff
@@ -372,6 +375,10 @@
               "$ref": "#/definitions/booleanRestrictions"
             }
           ]
+        },
+        "isArray": {
```
joneubank, line 379: Can remove from here now ;)

joneubank:
Reviewed too quickly... We don't want isArray to be a restriction, it should be where you had it originally.

PR #120, opened 2020-11-23
Rc/1.10.0 github container registry

blabadi:
This change adds a docker push to `ghcr` using the overturebio user account credentials that are already available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "lectern",
-  "version": "1.9.0",
+  "version": "1.10.0",
   "description": "Data Dictionary Management",
   "scripts": {
     "start": "npm run serve",


Jenkinsfile.groovy
```diff
@@ -1,5 +1,5 @@
 def commit = "UNKNOWN"
-
+def githubContainerRepo = "ghcr.io/overture-stack"
```
andricDu, line 2: I've personally named it dockerRepo as technically the pattern is the same regardless of registry.

andricDu (approved):

Jenkinsfile.groovy
blabadi: true, I just differentiated it because we still pushing to docker hub here too

blabadi: renamed

PR #126, opened 2021-01-06
[125] update vulnerable dependencies

caravinci:
Clears NPM Audit and updates NodeJS. 
Supersedes #123 and #124


README.md
```diff
@@ -18,7 +18,7 @@ The intended use of the data dictionaries is to describe the structure and valid
 ## Development
 
 ### Technology
-- Node.js 8.10+
+- Node.js 13.14+
```
caravinci, line 21: Outdated readme, doesn't match the node version used by Docker.

Dockerfile
```diff
@@ -1,10 +1,10 @@
-FROM node:12.5.0
+FROM node:13.14.0
```
caravinci, line 1: Updating to v.LTS-1, as we did with Persona

Dockerfile
```diff
@@ -1,10 +1,10 @@
-FROM node:12.5.0
```
blabadi, line 1: is this part of the vulnerability fix?

caravinci (commented):
Couple comments to help reviewing

Dockerfile
caravinci: I'm bringing it up to the later version we just updated Persona, so that devs don't have to switch back and forth when debugging multiple services.

Dockerfile
blabadi: I see, we had most new services (at least in argo) on 12 right now, because it was the active LTS version, my suggestion is if we upgrade we move to active LTS version (https://nodejs.org/en/about/releases/,  i.e. 14) 

devs could and probably should use nvm, but if this was a deliberate decision then cool with me we will upgrade later

caravinci: Indeed, it was a conscious decision to allow some "outdated environment" tolerance (hence `LTS - 1`), but I agree it'd be ideal to enforce LTS always. Do you suggest I change it to LTS in this PR?

blabadi: let's get this out and create ticket to upgrade to an LTS, I prefer major version changes to have their own releases to focus on vulnerability fixes

caravinci: Sounds splendid! 👌

blabadi (approved):

PR #151, opened 2021-12-07
Feat/#472 normalize new lines

ciaranschutte:
`fields: restrictions: script` can end up having either `\r\n\` or `\n` line endings.
This adds line ending normalising to `\n` for schema update, schema addition, dictionary addition

### DIFF
--- a/src/services/schemaService.ts
+++ b/src/services/schemaService.ts
@@ -34,19 +34,31 @@ export function validate(schema: any, references: any) {
   return { valid: validate(schemaWithReplacements), errors: validate.errors };
 }
 
+function normalizeScript(script: string | string[]) {
+  if (typeof script === 'string') {
+    return script.replace(/\r\n/g, '\n');
+  } else if (Array.isArray(script)) {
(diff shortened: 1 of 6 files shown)


blabadi (approved):
not sure why tests failed, didn't check, but the fix looks great to me
I assume existing schemas will be left as is this is to prevent future cases which is fine imo.

ciaranschutte:
> not sure why tests failed, didn't check, but the fix looks great to me I assume existing schemas will be left as is this is to prevent future cases which is fine imo.

Yes, I haven't touched existing schemas at all. Just a fix for future.
I will look into failing CI

src/services/schemaService.ts
```diff
@@ -33,3 +33,32 @@ export function validate(schema: any, references: any) {
 
   return { valid: validate(schemaWithReplacements), errors: validate.errors };
 }
+
+function normalizeScript(script: string | string[]) {
```
ciaranschutte, line 37: @blabadi modified to account for arrays. failing test caught it.

blabadi (approved):

PR #156, opened 2022-02-18
Security updates

ciaranschutte:
Will address most if not all dependabot alerts.

- major versions of mocha and gitmoji
- fixes all critical and high production production dependency vulnerabilities
- most dev dependency too. 
  - 2 remaining critical in `gitmoji` which is used to format git commit messages.
  - upgraded gitmoji to latest
 
 - type updates from updating packages (more details inline)


src/services/egoTokenService.ts
```diff
@@ -73,7 +73,7 @@ export default function(egoURL: string = process.env.EGO_API, scope: string = pr
         try {
           const scopes = authToken['context']['scope'] as Array<string>;
           if (scopes.includes(scope)) {
-            const routePromise = fn(req, res, next);
+            const routePromise: any = fn(req, res, next);
```
ciaranschutte, line 76: Unsure what type this could be

src/controllers/dictionaryController.ts
```diff
@@ -73,15 +74,15 @@ export const updateSchema = async (req: Request, res: Response) => {
...
+  const leftVersion = req.query.left.toString()
+  const rightVersion = req.query.right.toString();
 
   if (name && leftVersion && rightVersion) {
     const dict1Doc = await dictionaryService.findOne(name, leftVersion);
     const dict2Doc = await dictionaryService.findOne(name, rightVersion);
-    const dict1 = showReferences ? dict1Doc.toObject() : replaceReferences(dict1Doc.toObject());
-    const dict2 = showReferences ? dict2Doc.toObject() : replaceReferences(dict2Doc.toObject());
+    const dict1 = (showReferences ? dict1Doc.toObject() : replaceReferences(dict1Doc.toObject() as DictionaryDocument)) as DictionaryDocument;
+    const dict2 = (showReferences ? dict2Doc.toObject() : replaceReferences(dict2Doc.toObject() as DictionaryDocument)) as DictionaryDocument;
```
ciaranschutte, line 85: a little messy but gets the job done. `toObject` changed it's return type from `any` to `LeanDocument<DictionaryDocument>` which is a mongoose lighter version of a full document which has internal state and other additions that cause bloat for basic usage.

src/controllers/dictionaryController.ts
```diff
@@ -23,10 +23,11 @@ import { BadRequestError } from '../utils/errors';
 import { replaceReferences } from '../utils/references';
 import { diff as diffUtil } from '../diff/DictionaryDiff';
 import logger from '../config/logger';
+import { DictionaryDocument } from '../models/Dictionary';
 
 export const listDictionaries = async (req: Request, res: Response) => {
-  const name = req.query.name;
-  const version = req.query.version;
+  const name = req.query.name.toString();
+  const version = req.query.version.toString();
```
ciaranschutte, line 30: express added more typing `req.query` can be a string array or other specific types. using `.toString()` we always get a string then our actually functions can check if its a valid dictionary, version etc. Seems more logically to me than just casting to a string

ciaranschutte:
Will address most if not all dependabot alerts.

src/services/egoTokenService.ts
blabadi: I don't think it's a bid deal, but this is an express type `RequestHandler` example here https://github.com/overture-stack/ego-token-middleware/blob/master/src/index.ts#L2

blabadi:
NodeJS is where libraries like gitmoji have critical security threats, go figure

src/controllers/dictionaryController.ts
```diff
@@ -73,15 +74,15 @@ export const updateSchema = async (req: Request, res: Response) => {
 
 export const diffDictionaries = async (req: Request, res: Response) => {
   const showReferences = req.query.references || false;
-  const name = req.query.name;
-  const leftVersion = req.query.left;
-  const rightVersion = req.query.right;
+  const name = req.query.name.toString();
+  const leftVersion = req.query.left.toString()
+  const rightVersion = req.query.right.toString();
```
joneubank, line 79: Same question re: will this break if the client doesn't submit one of these params?

joneubank (commented):
Just one point of concern, otherwise looks good.

ciaranschutte:
@joneubank yeah I think I was trying to be too clever and you are right. I went back to a TS solution.. keep TS doing what its doing and not mess with the working logic.

ciaranschutte:
added node upgrade to solve build error
```
        throw Error(`yargs parser supports a minimum Node.js version of ${minNodeVersion}. Read our version support policy
        ```

**Review by joneubank — CHANGES_REQUESTED** [review]

File: src/controllers/dictionaryController.ts [hunk 5]
```diff
@@ -72,17 +79,28 @@ export const updateSchema = async (req: Request, res: Response) => {
...
+  req: Request<{}, {}, {}, { name: string; left: string; right: string; references: boolean }>,
+  res: Response,
+) => {
   const showReferences = req.query.references || false;
-  const name = req.query.name.toString();
-  const leftVersion = req.query.left.toString()
-  const rightVersion = req.query.right.toString();
+  const name = req.query.name;
+  const leftVersion = req.query.left;
+  const rightVersion = req.query.right;
```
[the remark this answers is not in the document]
ciaranschutte: not sure what you mean default Typescript query type.
If I use `Request` without specifying the 4th generic argument `findOne` complains.
At some point TS needs to know theyre strings and wont be array of query strings

I thought you might mean `Partial` so I have updated it with this. Just let me know if this was incorrect and I can fix. Thanks @joneubank

joneubank (approved):

Issue #177, opened 2023-06-22
Feature Request: Foreign Key restriction for a schema


joneubank:
Add a schema restriction that requires a value in records of this schema to have a matching value in a record on another schema. 

## Detailed Description
Each schema can be given a restriction that maps fields in this schema to fields in another schema (or other schemas). Every record validated by this schema will need to be validated against a data set for the other mapped schemas, the value in the specified field(s) in the record need to be found in a record on the other schema's data-set. If multiple foreign keys map to the same schema, then the values must be found in the same record.

As an example, a Specimen schema may add a foreignKey constraint mapping `specimen.donor_id` to the Donor schema field `donor.id`. Every record in specimen would have to find at least one record in the donor schema data-set that has a field `id` with a value matching the specimen's `donor_id`.

A foreign key restriction can include multiple fields and relations to multiple other schemas. This could be multiple fields to one schema, the same field mapped to multiple schema, or any such combination.

### A note on uniqueness
Lectern clients only need to validate that a single record exists in the mapped schema that matches the foreign key restriction - there is no restriction on this mapping being unique. If you require the values to be be unique in either schema then make use of the schema `uniqueKey` restriction https://github.com/overture-stack/lectern/issues/176 or restrict the individual fields to be unique https://github.com/overture-stack/lectern/issues/175.

## Possible Implementation
Add a `restrictions` object to the meta-schema for schemas, same one as described in https://github.com/overture-stack/lectern/issues/176 . 

Add to the `restrictions` object an optional property called `foreignKeys`. The value will be an array of foreign-schema mappings. 

Each of these mapping objects should include property `schema` with the string value of the name of the other schema, and a property `mappings` that will have a list of field mapping objects. 

The field mapping objects will have two properties, `local` with a string value that is the name of the local field that we are mapping from, and `foreign` which will have a string value that is the name of the field in the foreign schema that it will be matched to.

```json
{
  "restrictions": { 
    "foreignKeys":[
      {
        "schema": "donor", // <===== foreign schema name
        "mappings": [ // <===== can have multiple fields have foreign key mappings
          {
            "local": "donor_id", 
            "foreign": "id"
          }
        ]
      }
    ]
  }
}
```

Some code validations will be necessary in addition to the meta-data structure validation: 
* All `foreignKey.schema` values must be schema names that exist in the dictionary.
* All property names in the foreign key mappings must exist in their specified schemas.
[...]

PR #189, opened 2023-07-06
Add `foreignKey` schema restriction

joneubank:
Resolves: https://github.com/overture-stack/lectern/issues/177

Schemas can accept foreign key restriction with array of mappings. Validations are in place to make sure the local and foreign key names exist and are not self referential. Validations are also done to make sure that the types of the foreign and local fields are the same.


src/types/dictionaryTypes.ts
```diff
@@ -240,5 +292,62 @@ export const Dictionary = zod
...
+			{},
+		);
+
+		// Loop through each schema and apply checks if they have a foreignKey restriction
+		dictionary.schemas.forEach((schema) => {
+			if (schema.restrictions?.foreignKey) {
+				// We have a foreign key restriction, now we will now check its properties all reference existing schemas and fields
+
+				schema.restrictions.foreignKey.forEach((fkRestriction) => {
+					if (!schemaNames.includes(fkRestriction.schema)) {
```
justincorrigible, line 319: While I'd truly prefer you didn't use negative logic in general, in this particular case the extra cognitive burden is not just avoidable but unnecessary as you can just apply the opposite if.

justincorrigible (approved):

src/types/dictionaryTypes.ts
joneubank: Fair, I'll swap the order.

PR #190, opened 2023-07-11
Refactor Codebase into Monorepo

joneubank:
Introducing PNPM and Nx to make codebase a monorepo.

The long term objective is to allow us to introduce the Lectern Client code into the same repository so that they can share type definitions. This will also let us share the data validation code onto the Lectern Server, and the dictionary validation/diff code into the client.

At the end of this PR, lectern should build into the same server application, providing the same API. A further PR will add the Lectern Client code into this codebase.

## Approval Checklist
To accomplish the goal of this PR we need to confirm:
- [x] Jenkins is able to test and build the application from the monorepo
- [x] Jenkins is able to produce a working Dockerfile for the server module
- [x] Another user is able to clone the code base and run the server, following the instructions followed in the README
- [x] User Testing with [ARGO dictionary](https://github.com/icgc-argo/argo-dictionary) - a complicated use case of an active user of the application


.gitignore
```diff
@@ -1,64 +1,22 @@
...
-# next.js build output
-.next
+# build directories
+dist
 
-dist/
-.vscode
\ No newline at end of file
+# common editor application Configs
+.vscode
```
justincorrigible, line 21: I like leaving this un-ignored because it allows to suggest plugins, etc.

apps/server/package.json
```diff
@@ -0,0 +1,65 @@
+{
+	"name": "server",
```
justincorrigible, line 2: working on the assumption this gets appended by pnpm to the name in the root manifest?

apps/server/package.json
```diff
@@ -0,0 +1,65 @@
...
+	"version": "2.0.0-next.0",
+	"description": "Overture Data Dictionary Management",
+	"scripts": {
+		"start": "npm run serve",
+		"serve": "node dist/server.js",
```
justincorrigible, line 7: what's a `ts-node`?

apps/server/src/app.ts
```diff
@@ -27,7 +27,7 @@ import * as swagger from './config/swagger.json';
 import dictionaryRouter from './routers/dictionaryRouter';
 import diffRouter from './routers/diffRouter';
 import healthRouter from './routers/healthRouter';
-import { errorHandler } from './utils/errors';
+import { errorHandler } from './utils/errorHandler';
```
justincorrigible, line 30: 👍

apps/server/src/config/swagger.json
```diff
@@ -337,7 +337,7 @@
 				"in": "query",
 				"required": false,
 				"schema": {
-					"type": "string"
+					"type": "boolean"
```
justincorrigible, line 340: what the what? `schema: true`?
explain, for my own edification?

apps/server/src/db/dictionary.ts
```diff
@@ -82,5 +84,10 @@ export const listByName = async (name: string): Promise<DictionaryDocument[]> =>
...
 	const dict = new DictionaryModel(dictionary);
 	const saved = await dict.save();
-	return saved.toObject();
+	return stripExtras(saved.toObject());
 };
+
+/**
+ * Remove the mongo version property __v from the response object
+ */
+const stripExtras = (doc: DictionaryDocument) => omit(doc, '__v');
```
justincorrigible, line 93: went through a rabbit hole trying to figure out whether `stripExtras` is a native js thing, or a TS thing, or what...

do we want to move this definition to somewhere before it's implemented/executed to prevent that happening to other devs?

apps/server/src/external/ego.ts
```diff
@@ -18,11 +18,11 @@
  */
 
 import axios from 'axios';
+import { ForbiddenError, UnauthorizedError } from 'common';
```
justincorrigible, line 21: is `common` an npm package? looks like it used to be an internal import.
a bit confused about this change

apps/server/src/routers/diffRouter.ts
```diff
@@ -51,8 +50,9 @@ router.get(
 		const dict2Doc = await dictionaryService.getOneByNameAndVersion(name, rightVersion);
 		const dict1 = showReferences ? dict1Doc : replaceReferences(dict1Doc);
 		const dict2 = showReferences ? dict2Doc : replaceReferences(dict2Doc);
-		const diff = diffUtil(dict1, dict2);
-		res.status(200).send(Array.from(diff.entries()));
+		const diff = DiffUtils.getDiff(dict1, dict2);
+
+		res.status(200).send(DiffUtils.diffMapToArray(diff));
```
justincorrigible, line 55: Not a request for change, but a suggestion from experience, to prevent the occasional double response that's hard to troubleshoot.

```suggestion
		return res.status(200).send(DiffUtils.diffMapToArray(diff));
```

apps/server/src/routers/healthRouter.ts
```diff
@@ -21,7 +21,7 @@ import { dbHealth, Status } from '../app-health';
 
 const router = Router();
 
-router.get('/', (req, res) => {
+router.get('/', (_req, res) => {
```
justincorrigible, line 24: is this to imply `req` not being used? seems superfluous a change

apps/server/src/server.ts
```diff
@@ -17,16 +17,13 @@
...
 
-import { Server } from 'http';
 import mongoose from 'mongoose';
 import App from './app';
 import { setDBStatus, Status } from './app-health';
 import { getAppConfig } from './config/appConfig';
 import logger from './config/logger';
 import { constructMongoUri } from './utils/mongo';
 
-let server: Server;
```
justincorrigible, line 28: 🎉

apps/server/src/services/dictionaryService.ts
```diff
@@ -17,14 +17,14 @@
...
 import * as immer from 'immer';
+import { omit } from 'lodash';
 import logger from '../config/logger';
 import * as DictionaryRepo from '../db/dictionary';
 import { normalizeSchema, validate } from '../services/schemaService';
-import { Dictionary, Schema } from '../types/dictionaryTypes';
-import { BadRequestError, ConflictError, NotFoundError } from '../utils/errors';
-import { incrementMajor, incrementMinor, isGreater } from '../utils/version';
-import _ from 'lodash';
+import { VersionUtils } from 'dictionary';
```
justincorrigible, line 27: there is another import from `dictionary` in line 21, this could be consolidated there.
another one of those npm package imports I didn't know existed.
is this just a confusing "absolute" import syntax?

apps/server/src/config/swagger.json
joneubank: This just changes the input type in the swagger UI for the query parameter `references`

apps/server/src/routers/healthRouter.ts
joneubank: Indicating unused parameter. Gets flagged by TS because of the `noUnusedParamter=true` setting in .tsconfig . Helps indicate to a reader that the argument is a function placeholder and not used. Minor but I find it helpful :)

Issue #200, opened 2024-05-19
Feature Request: `meta` property should allow nested data


joneubank:
The `meta` property of dictionaries, schemas, and fields are currently restricted to being a flat key-property object. We have since found relevant use cases where we would like to be able to have nested data structures. We want to change the Lectern meta-schema to allow all `meta` properties to allow nested objects.

## Example Use Case

The following is a proposed partial field definition for a `date_of_birth` field in a proposed schema. There is a desire to indicate how this field maps to other ontologies or data structures. This is done with a `meta` property named `mappings`

In the current system, a complicated string with different delimiters is provided. It is fine for human reading but can't easily be used programatically without string parsing.
```json
{
    "name": "date_of_birth",
    "description": "Indicate participant's date of birth.",
    "valueType": "string",
    "meta": {
        "displayName": "Date Of Birth",
        "mappings": "All4One: Month,year of birth;mCODE STU3: Patient.birthDate;Phenopacket: Individual.date_of_birth;Beacon V2: Individual.date_of_birth",
        "source": "MOH"
    }
}
```

When meta fields are allowed to be nested we could instead represent this data as follows:
```json
{
      "name": "date_of_birth",
      "description": "Indicate participant's date of birth.",
      "valueType": "string",
      "meta": {
        "displayName": "Date Of Birth",
        "mappings": {
          "All4One": "Month,year of birth",
          "mCODE STU3": "Patient.birthDate",
          "Beacon V2": "Individual.date_of_birth"
        },
        "source": "MOH"
      }
}
```

## Detailed Description
- [ ] Update `meta` value type validations that are used in all levels of the Lectern meta schema (Dictionary, Schema, Field)
- [ ] Allow properties inside a `meta` value object to be a meta object.
  - This is recursive, so `meta` values have an unrestricted amount of nesting to work with

joneubank:
@edsu7 Can you review the example I used to justify this change? I took it from the example you wrote in the comment on the conditional restrictions draft PR. The updated version is how I imagine you would prefer to capture this data.

PR #201, opened 2024-05-19
Meta properties can be nested

joneubank:
Issue: https://github.com/overture-stack/lectern/issues/200

This change allows values inside `meta` properties to be nested objects. This done recursively so that any level of nesting is possible.

This is achieved by updating the zod schema for the DictionaryMeta type, which is re-used for all uses of meta properties, including in the Dictionary, in all Schemas, and in all Fields.

Additionally, the allowed types of meta values was extended to also allow arrays of strings, and arrays of numbers. Arrays of mixed types and arrays of booleans are still not allowed.

### DIFF
--- a/generated/DictionaryMetaSchema.json
+++ b/generated/DictionaryMetaSchema.json
@@ -38,10 +38,35 @@
     "Meta": {
       "type": "object",
       "additionalProperties": {
-        "type": [
-          "string",
-          "number",
-          "boolean"
+        "anyOf": [
+          {
+            "anyOf": [
+              {
+                "type": "string"
+              },
+              {
+                "type": "number"
+              },
+              {
+                "type": "boolean"
+              },
+              {
+                "type": "array",
+                "items": {
+                  "type": "string"
+                }
+              },
+              {
+                "type": "array",
+                "items": {
+                  "type": "number"
+                }
+              }
+            ]
+          },
+          {
+            "$ref": "#/definitions/Meta"
+          }
         ]
       }
     },
--- a/libraries/dictionary/src/types/dictionaryTypes.ts
+++ b/libraries/dictionary/src/types/dictionaryTypes.ts
@@ -29,7 +29,7 @@ export type NameString = zod.infer<typeof NameString>;
 
 export const Integer = zod.number().int();
 
-// Meta accepts as values only stirngs, numbers, booleans, arrays of numbers or arrays of strings
+// Meta accepts as values only strings, numbers, booleans, arrays of numbers or arrays of strings
 // Another Meta object can be nested inside a Meta property
 export const DictionaryMetaValue = zod.union([
 	zod.string(),
--- a/libraries/dictionary/src/types/dictionaryTypes.ts
+++ b/libraries/dictionary/src/types/dictionaryTypes.ts
@@ -29,11 +29,20 @@ export type NameString = zod.infer<typeof NameString>;
 
 export const Integer = zod.number().int();
 
-// Unlike references, meta is not nested and accepts as values only primitives or arrays of them.
-export const DictionaryMetaValue = zod.union([zod.string(), zod.number(), zod.boolean()]);
+// Meta accepts as values only stirngs, numbers, booleans, arrays of numbers or arrays of strings
+// Another Meta object can be nested inside a Meta property
+export const DictionaryMetaValue = zod.union([
+	zod.string(),
+	zod.number(),
+	zod.boolean(),
+	zod.array(zod.string()),
+	zod.array(zod.number()),
+]);
 export type DictionaryMetaValue = zod.infer<typeof DictionaryMetaValue>;
-export const DictionaryMeta = zod.record(DictionaryMetaValue);
-export type DictionaryMeta = zod.infer<typeof DictionaryMeta>;
+export type DictionaryMeta = { [key: string]: DictionaryMetaValue | DictionaryMeta };
+export const DictionaryMeta: zod.ZodType<DictionaryMeta> = zod.record(
+	zod.union([DictionaryMetaValue, zod.lazy(() => DictionaryMeta)]),
+);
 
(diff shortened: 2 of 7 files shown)


generated/DictionaryMetaSchema.json
```diff
@@ -38,10 +38,35 @@
...
+              {
+                "type": "array",
+                "items": {
+                  "type": "string"
+                }
+              },
+              {
+                "type": "array",
+                "items": {
+                  "type": "number"
```
justincorrigible, line 62: unclear why array of booleans are not needed/allowed.
 
worth commenting?

libraries/dictionary/src/types/dictionaryTypes.ts
```diff
@@ -29,11 +29,20 @@ export type NameString = zod.infer<typeof NameString>;
 
 export const Integer = zod.number().int();
 
-// Unlike references, meta is not nested and accepts as values only primitives or arrays of them.
-export const DictionaryMetaValue = zod.union([zod.string(), zod.number(), zod.boolean()]);
+// Meta accepts as values only stirngs, numbers, booleans, arrays of numbers or arrays of strings
```
justincorrigible, line 32:
```suggestion
// Meta accepts as values only strings, numbers, booleans, arrays of numbers or arrays of strings
```

libraries/dictionary/src/types/dictionaryTypes.ts
```diff
@@ -29,11 +29,20 @@ export type NameString = zod.infer<typeof NameString>;
...
+// Another Meta object can be nested inside a Meta property
+export const DictionaryMetaValue = zod.union([
+	zod.string(),
+	zod.number(),
+	zod.boolean(),
+	zod.array(zod.string()),
+	zod.array(zod.number()),
+]);
 export type DictionaryMetaValue = zod.infer<typeof DictionaryMetaValue>;
```
justincorrigible, line 41: this line reads confusingly, due to the identical capital capitalisation of type and variable to infer from.

maybe naming itself should be different? (e.g. additional description for each to differentiate one from the other)

libraries/dictionary/src/types/referenceTypes.ts
```diff
@@ -37,7 +37,7 @@ export type ReferenceArray = zod.infer<typeof ReferenceArray>;
 // References are recursive, but Zod can't do TS type inference for recursive definitions.
 // So in this one case we define the type first with the recursive structure and use it as a type-hint
 // for our zod schema. Reference: https://zod.dev/?id=recursive-types
-export type References = { [x: string]: ReferenceArray | ReferenceValue | References };
+export type References = { [key: string]: ReferenceArray | ReferenceValue | References };
```
justincorrigible, line 40: good!

libraries/dictionary/test/dictionaryTypes.spec.ts
```diff
@@ -471,4 +472,34 @@ describe('Dictionary Types', () => {
...
+		it('Can accept arrays of strings', () => {
+			const meta = { a: 'string', b: 123, c: true, array: ['asdf', 'qwerty'] };
+			expect(DictionaryMeta.safeParse(meta).success).true;
+		});
+		it('Can accept arrays of numbers', () => {
+			const meta = { a: 'string', b: 123, c: true, array: [123, 456, 789] };
+			expect(DictionaryMeta.safeParse(meta).success).true;
+		});
+		it('Cannot accept arrays of booleans', () => {
+			// This constraint feels a bit arbitrary but I can't imagine a clear use case for this.
```
justincorrigible, line 495: As commented before, it's unclear "why", so arbitrary or not, behooves explanation

justincorrigible (approved):
LGTM, yet left comments and questions to make it better (including typo correction)

generated/DictionaryMetaSchema.json
joneubank: this is a generated file. I can comment this point but not here.

libraries/dictionary/src/types/dictionaryTypes.ts
joneubank: This is a common recurring pattern with type schemas. We need to either rename all the schema objects to have names with the Schema suffix or accept that there are types and schemas with the same name. The latter is proposed in the zod documentation and is the pattern followed here.

libraries/dictionary/test/dictionaryTypes.spec.ts
joneubank: So my thinking here is: 

There are only two allowed values (true and false), so an array of booleans only communicates information based on the length of the array, and the position of each Boolean value in the array. If you build an array where the position of each boolean value matters, then that introduces a lot of ambiguity about the meaning of each position in the array, and it would be prone to typos where the values are in the wrong position. I don't want to enable that kind of ambiguity. If what you want is to represent a sequence where only the order matters, that is still accomplishable by a string or number array and those are not limited to Boolean states.

In the end though, if a good use case is presented we can change this decision.

libraries/dictionary/src/types/dictionaryTypes.ts
justincorrigible: in other words, limit the standards for the sake of a library... just like `import z` opinion?

questionable

joneubank: Difference here is I like this pattern. The schema validates for a type of the same name, it's quite intuitive to work with.

libraries/dictionary/test/dictionaryTypes.spec.ts
justincorrigible: that adds clarity, for sure. thanks.
is this rationale documented anywhere I'm not currently aware of?

libraries/dictionary/src/types/dictionaryTypes.ts
justincorrigible: that may be so for those without cognitive impairments, perhaps. 🤷‍♂️

PR was already approved anyway, declaritive enough, or not. ✅

PR #207, opened 2024-05-26
Move Lectern JS Client into monorepo

joneubank:
Adds the Lectern Client into the mono repo code base.


joneubank:
Closing to change merge order - Will merge the client code into develop directly instead of adding it off of the conditional-restrictions branch.

joneubank:
Replaced with: https://github.com/overture-stack/lectern/pull/208

Issue #206, opened 2024-05-26
Feature Request: Move Lectern JS Client into monorepo


joneubank:
We want to move the lectern client codebase into this repository so they can be maintained as part of the same monorepo.

Current js-lectern-client repo: https://github.com/overture-stack/js-lectern-client

There are many advantages to this move, but some important ones are:
- using the same type system between server and client
- enforcing changes to dictionary model are reflected in server and validation engine at same time
- simplifying documentation and user onboarding by gathering information about server and client in the same location

## Details
- [x] Create a new monorepo directory called `packages` to store packages published to npm
- [x] Create a new package called `client` to store the js-lectern-client code
- [x] Move existing lectern client code here

Once the client code is co-located with the server additional issues will be needed to track some improvements to the client from its current state
- [x] Update the types used by the client to reference the `dictionary` library
- [x] Move data validation into a shared library so it is usable on the server or in the client
- [x] Remove the generated HTML dev docs and standardize documentation between server and client
- [x] README documentation for devs, client functionality and how to use

Once the client has been published from this repository:
- [ ] Deprecate js-lectern-client repository

joneubank:
Removed the task to make the client runnable in browser and moved that to its own issue: #213 . The changes to allow this relate to script validations and it is desired to remove these instead of modifying that code at all. Temporarily we will live with this non browser version.

joneubank:
Readme updated to describe current state: https://github.com/overture-stack/lectern/commit/6784f6c74812ac3f7627ac53d06679088e57f0bf

joneubank:
Closing this ticket with the client merged and available in current state. Future work is needed to make it usable in the browser.

Other tasks that should be considered:
- Export all lectern dictionary content in the client, writing lectern types and parsing dictionaries is behaviour I would want from the client.
- Expose validation functions, currently only processing is available
- Rename the processing functions
- Separate processing into:
  - data structure validation
  - data type conversion
  - data value validation

PR #208, opened 2024-05-26
Add Lectern Client to monorepo

joneubank:
This adds the existing Lectern Client code into this monorepo. Previously this was written and maintained in [overture-stack/js-lectern-client](https://github.com/overture-stack/js-lectern-client). This code will now be maintained in the same directory as the Lectern server so that types and utilities can be shared.

The code added in this request includes a mostly unchanged copy/paste of the lectern client in its most up to date form. A small list of changes were made to make it align with the processes of this repository, including:
- Updating to latest TS version
- Update tsconfig to match other packages in monorepo
- Formatting with prettier to update indents from spaces to tabs
- Cleaning up build scripts and removing old publishing notes
- Move build directory from `/lib` to `/dist`
- Updating copyright years

This accomplishes the base goals of: #206 , but does not touch the refactoring tasks such as updating the client to use the same types as the server.


libraries/dictionary/package.json
```diff
@@ -5,7 +5,7 @@
 	"main": "dist/index.js",
 	"scripts": {
 		"build": "pnpm build:clean && tsc",
-		"build:clean": "rm -rf dist/ && mkdir dist",
+		"build:clean": "rimraf -rf dist/ && mkdir dist",
```
leoraba, line 8: no need arguments `-rf` when using rimraf

libraries/common/package.json
```diff
@@ -5,9 +5,12 @@
 	"main": "dist/index.js",
 	"scripts": {
 		"build": "pnpm build:clean && tsc",
-		"build:clean": "rm -rf dist/ && mkdir dist"
+		"build:clean": "rimraf -rf dist/ && mkdir dist"
```
leoraba, line 8: no need arguments -rf when using rimraf

leoraba (approved):
just left a comment overall LGTM

packages/client/package.json
```diff
@@ -0,0 +1,52 @@
+{
+	"name": "@overturebio-stack/lectern-client",
```
joneubank, line 2: Need to rename to `@overture-stack/lectern-client`. Once published, we should publish a deprecation notice with a link to the new package to the [original package location](https://www.npmjs.com/package/@overturebio-stack/lectern-client)

PR #210, opened 2024-05-27
Remove `script` restrictions

joneubank:
Removes all script restriction functionality from the lectern dictionary meta schema and from the client validation code.

All code related to managing scripts, such as the linebreak normalization code, is also removed.


joneubank:
Closing this request though we may review this later. We are reviewing how to phase out scripts with the introduction of conditional restrictions to see if we should create an intermediate version of Lectern with both types of restrictions available. This is to encourage incremental adoption by organizations currently using script validations which will need to be removed for future versions of Lectern dictionaries.

PR #216, opened 2024-07-23
Update property names in RecordError and TestResult types

joneubank:
## Objective

Update the names of two properties that are commonly used and were ambiguous. This change is incorporating feedback from a review of the validation-refactor branch.

Two properties are changed, then all uses of those properties throughout the validation and client packages are updated to reflect this change:

- `TestResult` - The data returned with Invalid test results is now found under the property `details` instead of `info`
- `FieldDetails` - The shared data type of all `RecordValidationError` objects has the `value` property renamed to `fieldValue`

### DIFF
--- a/packages/validation/src/types/testResult.ts
+++ b/packages/validation/src/types/testResult.ts
@@ -24,7 +24,7 @@ export const valid = (): TestResultValid => ({ valid: true });
 
 export type TestResultInvalid<T> = {
 	valid: false;
-	info: T;
+	details: T;
 };
 
 /**
@@ -35,7 +35,7 @@ export type TestResultInvalid<T> = {
  */
 export const invalid = <InvalidInfo>(info: InvalidInfo): TestResultInvalid<InvalidInfo> => ({
 	valid: false,
-	info,
+	details: info,
 });
 
 /**
--- a/packages/validation/src/validateRecord/RecordValidationError.ts
+++ b/packages/validation/src/validateRecord/RecordValidationError.ts
@@ -25,7 +25,7 @@ import type {
 
 export type FieldDetails = {
 	fieldName: string;
-	value: DataRecordValue;
+	fieldValue: DataRecordValue;
 };
 
 export type RecordValidationErrorInvalidValue = FieldDetails & FieldValidationErrorValueType;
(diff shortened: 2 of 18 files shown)


packages/validation/src/types/testResult.ts
```diff
@@ -24,7 +24,7 @@ export const valid = (): TestResultValid => ({ valid: true });
 
 export type TestResultInvalid<T> = {
 	valid: false;
-	info: T;
+	details: T;
```
joneubank, line 27: Property change for the `TestResult` invalid data.

packages/validation/src/validateRecord/RecordValidationError.ts
```diff
@@ -25,7 +25,7 @@ import type {
 
 export type FieldDetails = {
 	fieldName: string;
-	value: DataRecordValue;
+	fieldValue: DataRecordValue;
```
joneubank, line 28: Property name change for the `RecordValidationError` shared type `FieldDetails`.

packages/validation/test/validateDictionary/validateDictionary.spec.ts
```diff
@@ -65,9 +65,9 @@ describe('Dictionary - validateDictionary', () => {
 			expect(result.valid).false;
 			assert(result.valid === false);
 
-			expect(result.info.length, 'Only one schema found invalid.').equal(1);
-			expect(result.info[0]?.reason).equal('UNRECOGNIZED_SCHEMA');
-			expect(result.info[0]?.schemaName, 'Correctly names the unrecognized schema.').equal(wrongSchemaName);
+			expect(result.details.length, 'Only one schema found invalid.').equal(1);
+			expect(result.details[0]?.reason).equal('UNRECOGNIZED_SCHEMA');
```
justincorrigible, line 69: Is this worth adding conditional chaining here for this and all similar lines?

e.g.
```suggestion
			expect(result.details?.[0]?.reason).equal('UNRECOGNIZED_SCHEMA');
```

joneubank: The `assert` on line 66 informs us that we have the invalid test result, so we know we will have the `details` property.

PR #219, opened 2024-07-26
Rename all `convertValues` functions as `parseValues`

joneubank:
This improves naming clarity by describing the first stage in data processing as "Parsing" values instead of naming this process "Converting" values. The distinction is that the values are not being modified, they are simply being parsed from strings into the types defined in the schema.


justincorrigible (approved):
LGTM

PR #220, opened 2024-07-31
Consolidate Common and Dictionary packages

joneubank:
Combine the `common` package with the `dictionary` package.

Then this updates the dictionary package name to belong to the `@overture-stack` org. This is in preparation for publishing the Lectern client and its dependencies.

Screenshot 2024-07-31 at 3 04 25 PM

## Updates Included
- Packages that won't be published have now been set to private, specifically the server app and the scripts package
- README at the root level is updated to match the removed package and new package name
- All dependencies have been updated to make them consistent across packages. In particular the zod version is now consistent.


apps/server/package.json
```diff
@@ -1,5 +1,6 @@
 {
-	"name": "server",
+	"name": "@overture-stack/lectern-server",
+	"private": true,
```
joneubank, line 3: Updated name to include organization. Set to private.

packages/dictionary/package.json
```diff
@@ -1,6 +1,6 @@
 {
-	"name": "dictionary",
-	"version": "1.0.0",
+	"name": "@overture-stack/lectern-dictionary",
```
joneubank, line 2: This will be a publicly available dependency of both the client and server. Now set to be published under `@overture-stack` organization.

scripts/package.json
```diff
@@ -0,0 +1,16 @@
+{
+	"name": "lectern-scripts",
```
justincorrigible, line 2: should this also be scoped within the org?
we oughtn’t pollute the registry for others

justincorrigible (approved):
Le Squash and le Merge, l’ami 🧐
(if develop is the right target. thought you were aiming for a feature branch?)

joneubank:
Will be maintaining a feature branch for the conditional restriction feature which have breaking changes for the meta schema. This one only affects internal dependency organization that aren't yet published anywhere so there shouldn't be any risk in updating to develop.

joneubank:
Closing this PR as these changes were made before significant refactoring of the code base. The relevant changes will be replicated into a new feature branch based on the latest develop branch.

PR #221, opened 2024-08-01
Conditional Restrictions - Restrictions as Array, Unique as field property, No more scripts

joneubank:
## Objective

The primary purpose of this change is to allow fields to define restrictions as an array of restriction objects. This is a necessary step in order to allow multiple conditional restrictions.

To accomplish this, we need to separate the `unique` property from the restriction array as it needs to apply to every record in a schema to make sense - ie. it can't be part of a conditional restriction, it must be found at a consistent path within each field.

This also presented a good opportunity to remove script restrictions which need to be removed eventually anyways, since there is no test for scripts in the updated client, and there is no desire to rewrite the script normalization code to handle the restriction arrays.

## Details
- All fields allow restrictions to be a single restrictions object or an array of them
- `restrictions` property is still optional for all fields
- field validation `resolveRestrictions` function is updated to collect restriction rules from all objects in a restriction array (if it is an array)
- `unique` is now a property of a field instead of a restriction; it has been removed as a property of restriction objects
- `script` has been removed as a restriction property from all field types. function normalization code that was in place to format script functions has been removed.

### Tests

Tests were added to check:
- lectern-dictionary
  - fields are valid when restrictions is an arrays
  - fields are valid when restrictions is an object
  - fields are valid when restrictions is undefined
- lectern-validation
  - resolveRestrictions identifies restrictions from an array
  - field validation will apply multiple restrictions of the same type

## Tracking Meta-Schema Changes

This is the first of several breaking changes that will be made to the lectern dictionary meta-schema. All changes, breaking or additive, will be tracked in a new doc file `/docs/lectern-2.0-changes.md `.


packages/dictionary/src/metaSchema/dictionarySchemas.ts
```diff
@@ -113,38 +105,39 @@ export const SchemaFieldBase = zod
...
+		unique: zod.boolean().optional(),
 	})
 	.strict();
 export type SchemaFieldBase = zod.infer<typeof SchemaFieldBase>;
 
 export const SchemaStringField = SchemaFieldBase.merge(
 	zod.object({
 		valueType: zod.literal(SchemaFieldValueType.Values.string),
-		restrictions: StringFieldRestrictions.optional(),
+		restrictions: StringFieldRestrictions.or(StringFieldRestrictions.array()).optional(),
```
joneubank, line 116: Single or array of `StringFieldRestrictions` object.

apps/server/src/services/schemaService.ts
```diff
@@ -30,37 +32,3 @@ export function validate(schema: Schema, references: References): { valid: boole
...
-	const normalize = (script: string) => script.replace(/\r\n/g, '\n');
-
-	if (typeof input === 'string') {
-		return normalize(input);
-	} else {
-		return input.map(normalize);
-	}
-}
-
-export function normalizeSchema(schema: Schema): Schema {
```
joneubank, line 50: Only script fields were normalized. We have removed scripts from the Lectern Dictionary meta-schema in this PR.

packages/validation/src/validateField/resolveFieldRestrictions.ts
```diff
@@ -35,64 +67,5 @@ export const resolveFieldRestrictions = (
...
-			if (field.restrictions.required) {
-				output.push({ type: 'required', rule: field.restrictions.required });
-			}
-			// if (field.restrictions.script) {
-			// 	output.push({ type: 'script', rule: asArray(field.restrictions.script) });
-			// }
-			return output;
-		}
-	}
+	return TypeUtils.asArray(field.restrictions).flatMap(extractRulesFromRestrictionObject);
```
joneubank, line 70: Convert restrictions into an array if it was a single object, simplifies processing to always be done the same way.

ciaranschutte: woohoo native flatmap

apps/server/src/services/schemaService.ts
```diff
@@ -18,10 +18,12 @@
  */
 
 import { References, replaceSchemaReferences, Schema } from '@overture-stack/lectern-dictionary';
-import * as immer from 'immer';
```
ciaranschutte, line 21: ember anyone? ;)

packages/dictionary/src/utils/schemaUtils.ts
```diff
@@ -15,4 +18,6 @@ export const getRequiredFields = (schema: Schema): SchemaField[] =>
  * @returns
  */
 export const getOptionalFields = (schema: Schema): SchemaField[] =>
-	schema.fields.filter((field) => !field.restrictions?.required);
+	schema.fields.filter((field) =>
+		TypeUtils.asArray(field.restrictions).every((restrictionObject) => !restrictionObject?.required),
```
ciaranschutte, line 22:
```js
!undefined // true is ok?
```

packages/validation/test/fixtures/fields/multipleRestrictions/fieldStringArrayMultipleRegex.ts
```diff
@@ -0,0 +1,22 @@
...
+	},
+	restrictions: [
+		{
+			regex: regexRepeatedText,
+		},
+		{
+			regex: regexAlphaOnly,
+		},
+	],
+} as const satisfies SchemaStringField;
```
ciaranschutte, line 22: 👍

ciaranschutte (approved):

PR #222, opened 2024-08-04
Adds conditional restrictions to dictionary restriction schema

joneubank:
## Objective
Introduces Conditional Restrictions to the Lectern meta schema. All field types can now optionally contain conditions for applying restrictions based on the value of other fields in the same record.

The validation library has been updated to consider any conditional restrictions when resolving which restrictions need to be applied to a field.

### Dictionary
- Conditional Restrictions added to restrictions object type for all fields types (string/number/integer/boolean)
- Conditional Restriction match rules added:
  - `codeList`
  - `count`
  - `exists`
  - `range`
  - `regex`
  - `value`
- Reference replacement will now properly replace reference tags in:
  - recursively through nested meta objects
  - conditional restrictions, including nested recursive restrictions
  - `codeList` and `regex` match rules
- Formally removes `script` restrictions
- Moves `unique` restriction to be a property of a field, not part of restrictions
- Tests added:
  - recursive reference replacement
  - ensure conditional restrictions are parsed correctly by meta schemas

### Validation
- Conditional restrictions are resolved based on value of fields in record
- Tests added:
  - conditional restriction test match rules
  - conditional restriction array case matching

### Scripts
- JSON Schema generation script will now create a reference definition for field each field and each field's restriction types.
- New JSON Schema definition generated for the Lectern Dictionary, including conditional restriction types.


packages/dictionary/src/metaSchema/dictionarySchemas.ts
```diff
@@ -66,14 +66,23 @@ export type SchemaFieldValueType = zod.infer<typeof SchemaFieldValueType>;
 /* ****************************** *
  * Field Type Restriction Objects *
  * ****************************** */
-export const StringFieldRestrictions = zod
+export const BooleanFieldRestrictions = zod.object({ required: zod.boolean() }).partial();
```
joneubank, line 69: This section looks heavily modified but its just organizing the Restriction schemas alphabetically.

packages/dictionary/src/metaSchema/dictionarySchemas.ts
```diff
@@ -66,35 +67,112 @@ export type SchemaFieldValueType = zod.infer<typeof SchemaFieldValueType>;
...
+	.object({
+		if: ConditionalRestrictionTest,
+		then: IntegerFieldRestrictions.or(zod.lazy(() => IntegerFieldConditionalRestriction))
+			.or(zod.array(zod.union([IntegerFieldRestrictions, zod.lazy(() => IntegerFieldConditionalRestriction)])))
+			.optional(),
+		else: IntegerFieldRestrictions.or(zod.lazy(() => IntegerFieldConditionalRestriction))
+			.or(zod.array(zod.union([IntegerFieldRestrictions, zod.lazy(() => IntegerFieldConditionalRestriction)])))
+			.optional(),
+	})
+	.strict();
```
joneubank, line 112: This is how the conditional restrictions are defined. They use `zod.lazy()` to reference themselves since this is a recursive type definition... you can have a conditional restriction inside the `then` or `else` property of the condition. A similar object is defined for each field type (string/number/integer/boolean).

packages/dictionary/src/metaSchema/dictionarySchemas.ts
```diff
@@ -110,37 +188,37 @@ export const SchemaFieldBase = zod
...
+		restrictions: NumberFieldRestrictionsObject.or(NumberFieldRestrictionsObject.array()).optional(),
 	}),
 ).strict();
 export type SchemaNumberField = zod.infer<typeof SchemaNumberField>;
 
 export const SchemaIntegerField = SchemaFieldBase.merge(
 	zod.object({
 		valueType: zod.literal(SchemaFieldValueType.Values.integer),
-		restrictions: IntegerFieldRestrictions.or(IntegerFieldRestrictions.array()).optional(),
+		restrictions: IntegerFieldRestrictionsObject.or(IntegerFieldRestrictionsObject.array()).optional(),
```
joneubank, line 210: `IntegerFieldRestrictionsObject` could be the `IntegerFieldRestrictions` or `IntegerFieldConditionalRestriction` which is the new conditional restriction object we just defined.

packages/dictionary/src/metaSchema/restrictionsSchemas.ts
```diff
@@ -111,4 +114,99 @@ export const RestrictionRegex = zod.string().superRefine((data, context) => {
...
+	if: ConditionalRestrictionTest;
+	then?:
+		| TRestrictionObject
+		| ConditionalRestriction<TRestrictionObject>
+		| (TRestrictionObject | ConditionalRestriction<TRestrictionObject>)[];
+	else?:
+		| TRestrictionObject
+		| ConditionalRestriction<TRestrictionObject>
+		| (TRestrictionObject | ConditionalRestriction<TRestrictionObject>)[];
+};
```
joneubank, line 212: This is the shape of a ConditionalRestriction which all the specific implementations must adhere to.

apps/server/test/integration/fixtures/updateNewFile.json
```diff
@@ -30,7 +30,6 @@
 				"units": "days"
 			},
 			"restrictions": {
-				"script": ["validateWithMagic(check dependece on another field)"],
```
justincorrigible, line 33: I take it this is not necessary after the refactor?

packages/dictionary/test/fixtures/references/codeList_references/input.ts
```diff
@@ -35,14 +35,17 @@ const content: Dictionary = {
...
+						codeList: '#/SINGLE_VALUE',
+					},
 				},
 			],
 		},
 	],
 	references: {
 		ID_REG_EXP: '^[\\w\\s\\W]{5,}$',
 		SEX: ['Male', 'Female'],
+		SINGLE_VALUE: 'this',
```
joneubank, line 48: Used for a test ensuring that single values are wrapped in an array for codeList restrictions.

docs/lectern-2.0-changes.md
```diff
@@ -21,6 +22,7 @@ The release of Lectern 2.0 brings some important upgrades to the Lectern service
   - Updated interface for Lectern Server REST client
   - Exposes dictionary meta-schema validation, data parsing, and data validation functions
 
+
```
justincorrigible, line 25: probable mistake?
```suggestion
```

packages/dictionary/src/metaSchema/dictionarySchemas.ts
```diff
@@ -17,17 +17,18 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import { z as zod } from 'zod';
+import { z as zod, type ZodType } from 'zod';
```
justincorrigible, line 20: COOL! hadn't seen this done that way

packages/validation/src/validateField/restrictions/resolveFieldRestrictions.ts
```diff
@@ -0,0 +1,100 @@
...
+	const output = TypeUtils.asArray(restrictions).flatMap<FieldRestrictionRule>((restrictionObject) => {
+		if ('if' in restrictionObject) {
+			// This object is a conditional restriction, we will test the record vs the conditions then extract rules
+			//  from either the `then` or `else` block.
+			const result = testConditionalRestriction(restrictionObject.if, value, record);
+			if (result) {
+				return recursiveResolveRestrictions(restrictionObject.then, value, record);
+			} else {
+				return recursiveResolveRestrictions(restrictionObject.else, value, record);
+			}
```
joneubank, line 81: This is the code that detects conditional restrictions (by the `if` property) and then applies the logic to test their conditions and resolve restrictions based on `then` and `else` properties.

The rest of this file is not new, it was moved into this dir (`validateField/restrictions/resolveFieldRestrictions.ts`) from `validateField/resolveFieldRestrictions.ts`.

packages/dictionary/src/utils/typeUtils.ts
```diff
@@ -26,6 +26,44 @@
...
 
+/**
+ * Given a predicate function that checks for type `T`, this will create a new predicate funcion that
+ * will check if a value is of type `T[]`.
+ *
+ * @example
+ * // Create type and predicate for `Person`:
+ * type Person = { name: string; age: number };
+ * const isPerson = (value: unknown): value is Person =>
+ * 	!!value &&
```
justincorrigible, line 37: nitpicking on the example, for the sake of documentation simplicity (this line adds superfluous cognitive burden for a mere example)... type of falsey values will never be 'object', the first check, and therefor `!Person`.

```suggestion
```

packages/validation/src/validateField/conditions/testConditionalRestriction.ts
```diff
@@ -0,0 +1,130 @@
...
+): boolean => {
+	const fieldTestResults = values.map((value) => matchTest(value));
+	return resultForArrayTestCase(fieldTestResults, arrayCase || ARRAY_TEST_CASE_DEFAULT);
+};
+
+const testConditionForSingularValue = (
+	condition: RestrictionCondition,
+	_value: SingleDataValue,
+	fieldValues: DataRecordValue[],
+): boolean => {
```
joneubank, line 60: This is the function that tests a condition to see if which restrictions to apply.

The specific code for each of the matching cases are in dedicated files in this same dir.

The `_value` parameter here is the value of the field being tested. It is here in case needed for future use cases such as comparing its value vs another field to determine if restrictions are needed. This is likely to be removed in the next iteration. Since this function is not exported there will be no impact on the published library when changing this function definition.

scripts/src/generateMetaSchema.ts
```diff
@@ -22,11 +30,20 @@ console.log('Generating JSON Schema Meta-Schema...');
...
+		StringFieldRestrictions,
+		ConditionalRestrictionTest,
 		Meta: DictionaryMeta,
 		Name: NameValue,
+		ReferenceArray,
+		ReferenceTag,
+		References,
 		Schema,
 		SchemaField,
 	},
```
joneubank, line 49: This list of schemas is the list of reusable definitions that the JSON Schema will be able to reference by name using the `$ref` syntax.

packages/dictionary/src/metaSchema/dictionarySchemas.ts
joneubank: Yeah its nicer like this. You can set VSCode to automatically write type imports like this when it is adding import statements for you.

packages/dictionary/src/utils/typeUtils.ts
```diff
@@ -38,3 +76,42 @@ export const asArray = <T>(val: T | T[]): T[] => (Array.isArray(val) ? val : [va
...
+
+/**
+ * Determines if a variable is a number, with added restriction that it is Finite.
+ * This eliminates the values `NaN` and `Infinity`.
+ *
+ * Note: this is just a wrapper on `Number.isFinite` which is used by Lectern for identifying numbers in data value type checks.
+ * @param value
+ * @returns
+ */
+export const isNumber = (value: unknown): value is number => Number.isFinite(value);
```
justincorrigible, line 88: Nice!

docs/lectern-2.0-changes.md
joneubank: WIP document... capturing list of changes as bullet points for now so all changes are collected in one place and can be thoroughly described at a later point. Note on `placeholder` content as well.

packages/dictionary/src/metaSchema/dictionarySchemas.ts
justincorrigible: awesome! gotta share the how sometime this week over zoom/slack

packages/dictionary/src/utils/typeUtils.ts
joneubank: `null` is falsey and is type `object`

docs/lectern-2.0-changes.md
justincorrigible: 🤷 inconsistent double whitespace. if WIP ignore comment anyway 👍

apps/server/test/integration/fixtures/updateNewFile.json
joneubank: It was in fact breaking tests after I removed script restrictions.

packages/dictionary/src/utils/typeUtils.ts
justincorrigible: fds! you're right!
in that case precise intent expression is always preferable for its declarativeness
```suggestion
 * 	value != null &&
```

packages/dictionary/src/utils/typeUtils.ts
```diff
@@ -38,3 +76,42 @@ export const asArray = <T>(val: T | T[]): T[] => (Array.isArray(val) ? val : [va
...
+ * @returns
+ */
+export const isInteger = (value: unknown): value is number => Number.isInteger(value);
+
+/**
+ * Determines if a variables is of type number[], with add restriction that every element is an Integer.
+ * @param value
+ * @returns
+ */
+export const isIntegerArray = isArrayOf(isInteger);
```
justincorrigible, line 117: must be frank: though I get why this works, and it does well, I don't love how that reads out loud and that "smells faintly".

packages/dictionary/src/utils/typeUtils.ts
joneubank: Yeah thats probably clearer. I'll start using that instead.

packages/validation/src/parseValues/parseValues.ts
```diff
@@ -22,6 +22,7 @@ import {
 	failure,
 	failWith,
 	success,
+	TypeUtils,
```
justincorrigible, line 25: 🙌

packages/validation/src/utils/isValidValueType.ts
```diff
@@ -17,8 +17,8 @@
  * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-import type { DataRecordValue, SchemaField } from '@overture-stack/lectern-dictionary';
-import { isBooleanArray, isInteger, isIntegerArray, isNumber, isNumberArray, isStringArray } from './typeUtils';
+import { type DataRecordValue, type SchemaField, TypeUtils } from '@overture-stack/lectern-dictionary';
```
justincorrigible, line 20: man, I really like this! lol

apps/server/test/integration/fixtures/updateNewFile.json
justincorrigible: ah! that's great! good riddance

packages/dictionary/src/metaSchema/referenceSchemas.ts
```diff
@@ -23,7 +23,7 @@ export const ReferenceTag = zod
 	.string()
 	.regex(
 		RegExp('^#(/[-_A-Za-z0-9]+)+$'),
```
leoraba, line 25: should we scape forward slash?

joneubank: There are tests that show this is working. JS strings don't need to escape forward slash, and I don't think thats needed in regex either.

joneubank: Tests like this one: https://github.com/overture-stack/lectern/blob/develop/packages/dictionary/test/references.spec.ts#L53

Which references the following dictionary json with a reference tag: https://github.com/overture-stack/lectern/blob/develop/packages/dictionary/test/fixtures/references/simple_references/input.ts#L20

docs/dictionary-reference.md
```diff
@@ -170,15 +170,15 @@ A requirement condition is defined by providing a field name or list of field na
...
+| `value`                                                                                                                                                                                                                                               | all                       | Type of specified fields                           | Field value matches the value of the specified field. Strings are matched case insensitive. When arrays are matched, the order of their                                                                                                                                                                                                                                                                                        |
+| elements is ignored - a field matches this condition if the elements in field are the same elements as in the value match rule. For example, the rule `['abc', 'def']` matches the value `['def', 'abc']` but does not match `['abc', 'def', 'ghi']`. | `some_value`, `[1, 2, 3]` |
```
leoraba, line 181: this table is broken.

packages/validation/src/validateField/conditions/testConditionalRestriction.ts
```diff
@@ -0,0 +1,130 @@
...
+	}
+	return true;
+};
+
+const testConditionForArray = (
+	_condition: RestrictionCondition,
+	_value: ArrayDataValue,
+	_fieldValues: DataRecordValue[],
+): boolean => {
+	throw new Error('Unimplemented.');
```
leoraba, line 109: what would be the case when this function returns on error? when raw data comes as an array format `['apple', 'strawberry']` or when is comma separated `apple,strawberry`?

packages/validation/src/validateField/restrictions/testEmpty.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Test for required value on an array field. Before using the common pattern of applying the value test to
+ * each item in the array, we first check:
+ * - if the rule is `false` then the value is always valid
+ * - if the length of the array is 0 then the value is invalid
+ * @param rule
+ * @param values
+ * @returns
+ */
+const testEmptyArray = (rule: boolean, values: ArrayDataValue): TestResult<RestrictionTestInvalidInfo> => {
+	// Note: This doesn't apply the
```
leoraba, line 60: the rule?

ciaranschutte (approved):

packages/validation/src/validateField/conditions/testConditionalRestriction.ts
joneubank: The original values from a tsv would all be strings like `"apple,strawberry`". There are two steps to process the data, `parse` and `validate`. During parsing the value would be converted to `['apple', 'strawberry']`.

PR #223, opened 2024-08-19
Conditional Restrictions - Add conditional restriction logic to field meta schemas and implement validation logic

joneubank:
## Objective
Introduces Conditional Restrictions to the Lectern meta schema. All field types can now optionally contain conditions for applying restrictions based on the value of other fields in the same record.

The validation library has been updated to consider any conditional restrictions when resolving which restrictions need to be applied to a field.

### Dictionary
- Conditional Restrictions added to restrictions object type for all fields types (string/number/integer/boolean)
- Conditional Restriction match rules added:
  - `codeList`
  - `count`
  - `exists`
  - `range`
  - `regex`
  - `value`
- Reference replacement will now properly replace reference tags in:
  - recursively through nested meta objects
  - conditional restrictions, including nested recursive restrictions
  - `codeList` and `regex` match rules
- Formally removes `script` restrictions
- Moves `unique` restriction to be a property of a field, not part of restrictions
- Tests added:
  - recursive reference replacement
  - ensure conditional restrictions are parsed correctly by meta schemas

### Validation
- Conditional restrictions are resolved based on value of fields in record
- Tests added:
  - conditional restriction test match rules
  - conditional restriction array case matching

### Scripts
- JSON Schema generation script will now create a reference definition for field each field and each field's restriction types.
- New JSON Schema definition generated for the Lectern Dictionary, including conditional restriction types.

## Github Issue Tracking
This accomplishes the majority of tasks listed in #209 .
The missing tasks will be moved into separate tickets to be completed separately:
- `compare` restriction
- `compare` conditional match rule
- condition `arrayFieldCase` to modify the rules for matching values in a field with an array of values.


leoraba (approved):
🎉

Issue #204, opened 2024-05-24
Feature Request: Add Field meta schema property `delimiter`


joneubank:
When a field is an array, as indicated by the property `isArray`, it is assumed that when parsing a TSV all values for this field will use the `,` character as delimiter.

We will add an `delimiter` property that can provide a different character or string that will be used as the delimiter between values when parsing this field from a TSV.

## Rules
- fields may optionally have a `delimiter` value
- values must be a string and cannot be empty (min length 1), no max length (if a long delimiter is required that can be provided)
- when no value is provided the default used will be `,`

joneubank:
Updated the ticket description to change the default delimiter to be `,`

PR #227, opened 2024-08-30
Custom delimiter on array fields

joneubank:
Ticket: https://github.com/overture-stack/lectern/issues/204

## Objective
Adds the ability to define a custom delimiter on any field. If the field has an array value, the value will use this custom delimiter to split the string value into an array.

The property is named `delimiter` and accepts a string value. Delimiter values must be at least one character in length, but can be longer and can contain any combination of characters.

The [dictionary-reference](https://github.com/overture-stack/lectern/blob/feat/delimiter/docs/dictionary-reference.md#dictionary-field-structure) documentation has been updated to reflect the array delimiter implementation.

## Example

String array split on `|` character:

```ts
const exampleFieldWithCustomDelimiter: SchemaField = {
	name: 'pipe-separated-list',
	isArray: true,
	delimiter: '|',
	valueType: 'string'
};
```

Then when used to parse data by the function `parseFieldValue()` from the validation library:

```ts
const result = parseFieldValue('abc|def|ghi', exampleFieldWithCustomDelimiter);
// result.data === ['abc', 'def', 'ghi']
```


packages/validation/test/parseValues/parseField.spec.ts
```diff
@@ -300,5 +300,34 @@ describe('Parse Values - parseFieldValue', () => {
 			it('Boolean array field, rejects array where value is missing (two delimiters are adjacent)');
 			expect(parseFieldValue(',true,false,TRUE', fieldBooleanArrayRequired).success).false;
 		});
+		describe('Custom delimiter', () => {
```
justincorrigible, line 303: nice feature! "out loud" thoughts about this, not to question your coverage but to narrate my analysis: asking myself "how do you escape the delimiter if it's also meant to be used as a value?" like you would in a comprehensive array of symbols. the answer I'd guess would be "use a compound delimiter", and then I realised it's not clear what happens if I give the delimiter repeatedly. does the array end up with an empty value? or does the function filter empties?

e.g. assume default (i.e. `,`), what if I give you `symbols = ',",\,%,,,#,*` do I end up with `[' , " , \ , % , , , # , *].length === 8`? or `[' , " , \ , % , # , *].length === 6`
... and what would be the representation of the quotes in JS, since one of the two first elements here would have to be escaped to make a string.
There's special handling here that could use being documented, methinks?

docs/dictionary-reference.md
```diff
@@ -67,27 +67,27 @@ In addition to schemas, a Lectern Dictionary can contain reference values that c
...
+| `valueType`    | Required |                        | [Field Data Type](#field-data-types)                | Type of value stored in this field                                                                                                                                                                                                                                                                                                                                       | `"string"`                                           |
+| `delimiter`    | Optional | `,`                    | `string`                                            | Character or string that will be used to split multiple values into an array. The default delimiter is a comma `,`. Any characters can be used as a delimiter. The delimiter value can be one or more characters long, but cannot be an empty string. Note: This property has no effect unless the field has `isArray: true`.                                            | `"\|"`                                               |
```
demariadaniel, line 74: Just to double check and confirm, I assume there's validation and TS preventing a submission where `valueType: 'integer', isArray: true, delimiter: '0'` ?
isArray is true but valueType is not string

packages/validation/src/parseValues/parseValues.ts
```diff
@@ -138,7 +138,7 @@ const convertArrayValue = (value: string, fieldDefinition: SchemaField): Result<
 
 	/* === Start of convertArrayValue logic === */
 	const { valueType } = fieldDefinition;
-	const delimiter = DEFAULT_DELIMITER;
+	const delimiter = fieldDefinition.delimiter !== undefined ? fieldDefinition.delimiter : DEFAULT_DELIMITER;
```
justincorrigible, line 141: I think this would allow empty strings as a delimiter, which in effect is not a delimiter at all. am I missing something here?
possible fix/workaround:
```suggestion
	const delimiter = typeof fieldDefinition.delimiter === 'string' && fieldDefinition.delimiter.length ? fieldDefinition.delimiter : DEFAULT_DELIMITER;
```

packages/validation/test/parseValues/parseField.spec.ts
joneubank: if you put the delimiter back-to-back this results in an empty value between them. This is not allowed in arrays where each element is expected to have a value. Validation of the data will fail.

https://github.com/overture-stack/lectern/blob/1c32ea4b15233050f2e2383f7b107e4477874cb8/packages/validation/test/parseValues/parseField.spec.ts#L269-L271

joneubank: There is no handling for escaping characters. If you are providing data in a TSV, all characters are allowed. If you are 
providing data in JSON, then the JSON parsing will strip escape characters.

docs/dictionary-reference.md
joneubank: There is not. That is a valid delimiter, and it would prevent numeric values with `0` in them. I would advise against using this delimiter but the spec allows it atm.

packages/validation/src/parseValues/parseValues.ts
justincorrigible: Just realised zod schema states min 1... code here could still benefit from the additional declarativeness, though

justincorrigible (approved):
Left comment to improve declarativeness as well as triple ensure type safety for the custom delimiters 👍 

Nice feature!

Issue #213, opened 2024-06-18
Make the client usable in browser and server


joneubank:
The validation library and Lectern client are both meant to be runnable in browser or on a server. This will enable the creation of UIs that can validate data vs a Dictionary as users enter it in forms or without submitting data to a server. This will enable a lot of UI tooling.

This is currently not possible due to the use of Node specific libraries which cannot be used in browser. There are two libraries that need removing:
1. `winston` - used in the client logger
2. `vm` - used in the validation library to run script validation's in an isolated context

### Removing Winston - Logging
The logger implementation can be removed from the client. We can selectively include console logging.
If there are specific processes where we want logging to occur, we need ot make this configurable for the invoking software. First, we need to define a Logger interface that our server can implement. The client can then accept a Logger as an optional parameter. This configuration should also allow for specification of log level.

### Removing VM - Scripts
We need to remove script conditionals anyways, so the solution here is to not publish anymore versions of the Lectern Client that run script validations.

PR #228, opened 2024-09-06
fix to use lectern-client outside node env

ciaranschutte:
- use lectern-client in browser contexts
- use with modern build systems/frameworks
- needs Node 18 - replaces `node-fetch` with native fetch
- builds to es modules not commonjs
- conditionally enables winston logging based on env (checks for `process` global var)


joneubank:
https://github.com/overture-stack/lectern/issues/213

PR #229, opened 2024-09-18
export validation error type

leoraba:


joneubank:
I have no problem with this, but I wonder if we should also export more error details. A similar need could arise from the response to the other validation functions.

For example:
- `SchemaValidationRecordErrorDetails` to go with `SchemaRecordError`
- `FieldValidationErrorRestrictionInfo` to go with `FieldValidationError`
- `ParseRecordFailureData` to go with `ParseRecordResult`

joneubank (changes requested):
Let's be consistent and export error details from more of the responses. I listed the others here: https://github.com/overture-stack/lectern/pull/229#issuecomment-2361918030

leoraba:
Added additional types to export.

joneubank (approved):

PR #231, opened 2024-09-23
run integration tests separately

leoraba:
Separate integration tests that uses `testcontainers` due to CI Jenkins pipeline unable to connect to testcontainers instance

Create a new Script to run only the integration tests located at folder `test/integration/**/*`
Script command: `pnpm run integration-test:all`

The existing script `pnpm run test:all` will ignore folder `test/integration/**/*`


package.json
```diff
@@ -10,6 +10,7 @@
 		"build:server": "pnpm nx build @overture-stack/lectern-server",
 		"build:client": "pnpm nx build @overture-stack/lectern-client",
 		"test:all": "pnpm nx run-many --all --target=test",
+		"integration-test:all": "pnpm nx run-many --all --target=integration-test",
```
justincorrigible, line 13: what's the difference between "test" and "integration test"?
I think this change points out that the existing script naming (before this PR) needs some love.

leoraba: the "test" script will run all the test excluding the folder `test/integration/**/*` this exclusion is defined on the default .mocharc.json.

the "integration-test" scripts does the opposite, runs only the tests of the folder `test/integration/**/*`.  (defined on .mocharc.integration.json file

joneubank, line 13: How about using three scripts:
- test:unit -- runs unit tests
- test:integration -- runs integration tests, may include testcontainers
- test -- runs test:unit and test:integration. Can also have `test:all` which `test` would be an alias of.

joneubank (changes requested):

leoraba:
Renamed the script names to be more concise:

- "test:all": To run both Unit and Integration tests in all the projects
- "test:unit:all": To run Unit tests all the projects
- "test:unit:dictionary": To run unit tests in Dictionary package
- "test:unit:validation": To run unit tests in Validation package
- "test:unit:server": To run unit tests in Server package
- "test:unit:client": To run unit tests in client package
- "test:integration:server": To run unit tests in server package

package.json
leoraba: As there are multiple projects renamed scripts to include the package name. see [comment](https://github.com/overture-stack/lectern/pull/231#issuecomment-2377811701)

joneubank (approved):

joneubank:
Looks like we still need an update to the jenkins pipeline to accomodate this. Change the instruction to test:unit

PR #236, opened 2024-10-31
bug fix - nested restrictions

leoraba:
PR Changes include:
- Fix issue #237
- Export `ParseFieldError` type on Lectern client package. This type is used in parse functions.

### DIFF
--- a/packages/validation/test/fixtures/fields/conditionalRestrictions/fieldStringConditionalExistsWithouthThenElse.ts
+++ b/packages/validation/test/fixtures/fields/conditionalRestrictions/fieldStringConditionalExistsWithouthThenElse.ts
@@ -1,0 +1,27 @@
+import { SchemaField, type SchemaStringField } from '@overture-stack/lectern-dictionary';
+import { fieldStringNoRestriction } from '../noRestrictions/fieldStringNoRestriction';
+import { validateFixture } from '../../../testUtils/validateFixture';
+
+export const fieldStringConditionalExistsWithouthThenElse = {
+	name: 'conditional-field',
+	valueType: 'string',
+	description: 'Required if `fieldStringNoRestriction` field exists, otherwise must be empty',
+	restrictions: {
+		if: {
+			conditions: [
+				{
+					fields: [fieldStringNoRestriction.name],
+					match: {
+						exists: true,
+					},
+				},
+			],
+		},
+	},
+} as const satisfies SchemaStringField;
+
+validateFixture(
+	fieldStringConditionalExistsWithouthThenElse,
+	SchemaField,
+	'fieldStringConditionalExistsWithouthThenElse is not a valid SchemaField',
+);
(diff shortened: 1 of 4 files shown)


joneubank:
We should have a test to capture this issue. Since we don't can you add one?

packages/validation/test/parseValues/parseField.spec.ts
```diff
@@ -165,6 +167,34 @@ describe('Parse Values - parseFieldValue', () => {
 				expect(parseFieldValue('     !@#$%^&* ()_+      ', fieldStringNoRestriction).success).true;
 				expect(parseFieldValue('     !@#$%^&* ()_+      ', fieldStringNoRestriction).data).equals('!@#$%^&* ()_+');
 			});
+			it('Successfuly parses strings, with conditional restrictions', () => {
+				const value = 'any random string value!!!';
+				const result = parseFieldValue(value, fieldStringConditionalExists);
```
leoraba, line 172: testing parsing any string using a Schema Field Definition having `if/then/else` restrictions

packages/validation/test/parseValues/parseField.spec.ts
```diff
@@ -165,6 +167,34 @@ describe('Parse Values - parseFieldValue', () => {
...
+				expect(parseFieldValue('    123', fieldStringConditionalExists).success).true;
+				expect(parseFieldValue('    123', fieldStringConditionalExists).data).equals('123');
+				expect(parseFieldValue('false   ', fieldStringConditionalExists).success).true;
+				expect(parseFieldValue('false   ', fieldStringConditionalExists).data).equals('false');
+				expect(parseFieldValue('     !@#$%^&* ()_+      ', fieldStringConditionalExists).success).true;
+				expect(parseFieldValue('     !@#$%^&* ()_+      ', fieldStringConditionalExists).data).equals('!@#$%^&* ()_+');
+			});
+			it('Successfuly parses strings, with conditional restrictions without then or else', () => {
+				const value = 'any random string value!!!';
+				const result = parseFieldValue(value, fieldStringConditionalExistsWithouthThenElse);
```
leoraba, line 185: Testing parsing any string using a Schema Field Definition having an`if` restriction without `then/else`. This is not a functional Schema definition however to cover every scenario possible we have to test it.

packages/validation/test/fixtures/fields/conditionalRestrictions/fieldStringConditionalExistsWithouthThenElse.ts
joneubank, line 1: typo in file name: Withouth

packages/validation/test/parseValues/parseField.spec.ts
joneubank: Is the expected behaviour that an if without then/else just never applies restrictions?

Should we consider adding validation to lectern dictionaries to require an if to have a then or else (at least one of the two)?

joneubank (approved):

packages/validation/test/parseValues/parseField.spec.ts
leoraba: If not then/else is found in the schema definition the expected behaviour is do not apply the restriction. Based on test [resolveFieldRestrictions.spec.ts](https://github.com/overture-stack/lectern/blob/develop/packages/validation/test/validateField/resolveFieldRestrictions.spec.ts#L106)

leoraba: I'm open to add a validation to require at least one between then/else. Although it could be added in a separate PR

PR #247, opened 2025-05-20
Refactor lectern-client Rest Module to work in Browser or on Server

### LINKED ISSUE #213: Make the client usable in browser and server

joneubank:
## Description

This includes the final changes needed for the lectern-client to run inside a browser, while still running succesfully on the server.

This requires removing two dependencies that only work inside NodeJS (winston and node-fetch), and refactoring the modules that were using them. The client's rest module was re-written from scratch to provide functions to list all dictionaries on a server, and to fetch a single Dictionary either by ID or by Name and Version. 

Future work is needed to also include a getDiff or getSchema/getField. All the POST/PUT endpoints that modify data will also require auth integration, and these have been left for future tasks.

This is a breaking change in that it changes which fetch functions were available in the client, but it should also unblock any browser integrations.

### Server CORS
In addition to the client changes, the Lectern Server needed to have CORS configuration added so that requests from a browser would succeed. The server has the Express library updated to latest and the expressjs/cors middleware added. An environment variable is now provided that will allow configuration of the allowed CORS domains. By default, CORS will reject all domains, so no security changes for existing servers.

## Fixes 
- #213 

## Type of change

Please choose only the relevant options and remove the others.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

- Server has unit tests added for CORS functionality
- Client has been tested through REPL import of the built library and integration tests run against a local running Lectern Server to list and get dictionaries.

## Future Work
- Lyric uses the client, and will need to be updated to use the new rest method: https://github.com/overture-stack/lyric/blob/main/packages/data-provider/src/external/lecternClient.ts#L18

## Checklist:

You do not need to fullfill all requirements of this checklist, select all that apply:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings


apps/server/package.json
```diff
@@ -22,11 +22,15 @@
 	"homepage": "https://github.com/overture-stack/lectern#readme",
 	"devDependencies": {
 		"@types/body-parser": "^1.19.5",
```
ciaranschutte, line 24: remove types for unused `body-parser` package

apps/server/package.json
```diff
@@ -22,11 +22,15 @@
...
+		"@types/chai": "^4.3.16",
+		"@types/chai-http": "^4.2.4",
+		"@types/cors": "^2.8.18",
 		"@types/errorhandler": "0.0.32",
 		"@types/express": "^4.17.21",
```
ciaranschutte, line 29: upgraded `express` package but not it's types

packages/client/package.json
```diff
@@ -40,11 +40,10 @@
 	"dependencies": {
 		"@overture-stack/lectern-dictionary": "workspace:^",
 		"@overture-stack/lectern-validation": "workspace:^",
+		"axios": "^1.7.2",
```
ciaranschutte, line 43: comment.

web standard fetch is available in node
axios is overkill
aligns with removing things we don't need (`body-parser` removes in preference for native express api)

packages/client/src/changeAnalysis/changeAnalyzer.ts
```diff
@@ -29,10 +29,10 @@ const isNestedChange = (input: FieldChanges): input is NestedChanges => {
...
 };
 
-export const fetchDiffAndAnalyze = async (serviceUrl: string, name: string, fromVersion: string, toVersion: string) => {
-	const changes = await restClient.fetchDiff(serviceUrl, name, fromVersion, toVersion);
-	return analyzeChanges(changes);
-};
+// export const fetchDiffAndAnalyze = async (serviceUrl: string, name: string, fromVersion: string, toVersion: string) => {
+// 	// const changes = await restClient.fetchDiff(serviceUrl, name, fromVersion, toVersion);
+// 	return analyzeChanges(changes);
+// };
```
ciaranschutte, line 35: delete or add comment explaining why code is commented out

packages/client/README.md
```diff
@@ -23,18 +23,29 @@ The Lectern Client provides developers TypeScript code tools to interact with Le
...
-const dictionaryName = 'dictionary-name';
-const currentVersion = "2.3";
-const previousVersion = "2.1";
+const dictionaryName = 'my-example-schemas';
+const dictionaryVersion = "2.3";
+
+const allAvailableDictionariesResult = await lectern.rest.listDictionaries(lecternUrl);
+const filteredByNameDictionariesResult = await lectern.rest.listDictionaries(lecternUrl, {name: dictionaryName});
+
+const exampleDicionaryResult = await lectern.rest.fetchDiff(lecternUrl, {name: dictionaryName, version: dictionaryVersion});
```
ciaranschutte, line 38: is `fetchDiff` removed? description of PR says 

"Future work is needed to also include a getDiff or getSchema/getField. All the POST/PUT endpoints that modify data will also require auth integration, and these have been left for future tasks."

ciaranschutte (approved):
couple of comments. tests + validation are looking great.

apps/server/package.json
joneubank: This helped me spot an underlying type issue, thank you.

packages/client/package.json
justincorrigible: `fetch` in any of its flavors doesn't include cancelling (you'd have to instantiate an AbortController separately), which is something we've been meaning to start implementing consistently.

does that add any nuance to this choice?

apps/server/.env.example
```diff
@@ -15,6 +15,9 @@
 # AUTH_ENABLED=false
 # EGO_API=
 # SCOPE=
+# # CORS allowed origins can be a comma separated list of the allowed domains.
```
justincorrigible, line 18: does this parse spacing? e.g. `a,b,c` vs `a, b, c`

apps/server/README.md
```diff
@@ -36,13 +36,21 @@ docker-compose up -d
 To build the server, and all modules it depends on:
 
 ```shell
-pnpm nx build server
+pnpm nx build @overture-stack/lectern-server
+```
+or
+```shell
+pnpm build
```
justincorrigible, line 43: what's the difference, if any? otherwise, line 43 should suffice as "lower complexity" choice

apps/server/.env.example
joneubank: no, this takes whitespace as part of the value. very simple implementation

apps/server/README.md
```diff
@@ -36,13 +36,21 @@ docker-compose up -d
...
 
 To run the server:
 
 ```shell
-pnpm nx start server
+pnpm nx start @overture-stack/lectern-server
+```
+or
+```shell
+pnpm start
```
justincorrigible, line 53: same here. difference? otherwise, simpler is best

apps/server/README.md
```diff
@@ -87,25 +95,26 @@ The .env file you pass to this command can follow the formatting of the template
...
+| -------------------- | ------------------------------ | ------------ | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| OPENAPI_PATH         | No                             | String       | `/api-docs` | Path to Swagger UI with API documentation.                                                                                                              |
+| PORT                 | No                             | Number       | `3000`      | Port Lectern Server API will listen to.                                                                                                                 |
+| CORS_ALLOWED_ORIGINS | No                             | String[]     | -           | List of domains that will be allowed by CORS. Multiple domains can be listed, separated by commas. Example: `http://localhost:5173,https://example.com` |
```
justincorrigible, line 102: ah, looks like this is the answer to my earlier question. perhaps examples would be useful in the `.env.example`? (unless the intent was to create a `.env.template` instead)
```suggestion
| CORS_ALLOWED_ORIGINS | No                             | String[]     | -           | List of domains that will be allowed by CORS. Multiple domains can be listed, separated by commas. No spaces. Example: `http://localhost:5173,https://example.com` |
```

apps/server/src/config/appConfig.ts
```diff
@@ -66,6 +67,10 @@ const buildAppContext = async (secrets: any): Promise<AppConfig> => {
 			return process.env.OPENAPI_PATH || '/api-docs';
 		},
 
+		corsAllowedOrigins(): string[] {
+			return process.env.CORS_ALLOWED_ORIGINS ? process.env.CORS_ALLOWED_ORIGINS.split(',') : [];
```
justincorrigible, line 71: either the "no spaces" in readme, or this here?

```suggestion
			return process.env.CORS_ALLOWED_ORIGINS ? process.env.CORS_ALLOWED_ORIGINS.split(',').map(trim) : [];
```

joneubank: this is fine idea, I'll add it.

apps/server/src/routers/healthRouter.ts
```diff
@@ -27,13 +27,15 @@ router.get('/', (_req, res) => {
 			appStatus: 'Up',
 			dbStatus: dbHealth.status,
 		};
-		return res.status(200).send(resBody);
+		res.status(200).send(resBody);
+		return;
```
justincorrigible, line 31: naive question: return is implicit, what's the purpose of putting it separately from the response? surely `return res...` has the exact same effect without looking unnecessarily verbose.

apps/server/README.md
joneubank: I'll add the trim thing instead.

apps/server/src/routers/healthRouter.ts
justincorrigible: ugh Express! le why you le break? 😞 
it was a nice pattern! are you sure it's not just their TS implementation that's the problem?

PR #252, opened 2025-05-27
#218: add endpoint to download TSV template files for all data entities

Azher2Ali:
## Summary

Added a new API endpoint to Lectern that generates TSV template files for all data entities defined in a dictionary and bundles them into a downloadable zip file.

### fixes
- #218

## Description of Changes

- Added a new GET /templates/download endpoint.
- Endpoint accepts dictionaryName and dictionaryVersion as query parameters to download templates for a specific dictionary version.
- Generates a TSV template file for each entity defined in the dictionary’s base schema.
- Bundles all generated template files into a single zip archive and returns it as a downloadable response.
- Added utility functions to create TSV templates from entity definitions.
- Added a service method to gather all entity templates based on a given dictionary name and version.

## Readiness Checklist

- [ ] **Self Review**
- [ ] **Formatting**
- [ ] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation


apps/server/src/controllers/dictionaryController.ts
```diff
@@ -105,7 +107,7 @@ export const getSchema = async (req: Request<{ dictId: string; schemaName: strin
 	const dictionary = await dictionaryService.getOneById(dictId);
 	const formattedDictionary = replaceReferences(dictionary);
 
-	const schema = formattedDictionary.schemas.find((schema) => schema.name === schemaName);
+	const schema = formattedDictionary.schemas.find((schema: { name: string }) => schema.name === schemaName);
```
joneubank, line 110: Why is this type declaration necessary? Is it not inferring the type of schema from `formattedDictionary.schemas`?

apps/server/src/controllers/dictionaryController.ts
```diff
@@ -133,19 +135,56 @@ export const getSchemaField = async (
...
 };
+
+export const downloadTemplates = async (req: Request<{}, {}, {}, { name: string; version: string }>, res: Response) => {
+	const { name, version } = req.query;
+
+	if (!name || !version) {
+		throw new BadRequestError('Missing dictionary name or version in query.');
+	}
+
+	const dictionary = await dictionaryService.downloadDictionaryByNameAndVersion(name, version);
```
joneubank, line 162: Need to handle error cases where dictionary is not found, or if the DB fetch fails.

apps/server/src/controllers/dictionaryController.ts
```diff
@@ -133,19 +135,56 @@ export const getSchemaField = async (
...
+
+	for (const schema of dictionary.schemas) {
+		const fields = schema.fields || [];
+		const templateRow = fields.reduce(
+			(acc: { [x: string]: string }, field: { name: string | number }) => {
+				acc[field.name] = '';
+				return acc;
+			},
+			{} as Record<string, string>,
+		);
```
joneubank, line 174: What you have here is correct and functional, I personally prefer specifying the generic type of the reduce function instead of asserting the type of the initial accumulator value:

```ts
		const templateRow = fields.reduce<Record<string, string>>(
			(acc: { [x: string]: string }, field: { name: string | number }) => {
				acc[field.name] = '';
				return acc;
			},
			{},
		);
```

The advantage here (imo) is to see the type of the accumulator at the start of the function, and then also TS will validate if the initial accumulator value is valid for the accumulator type. This avoids a potential error case where the type assertion declares a type that the value does not match.

joneubank (commented):
Read this before all code was pushed, comments on the section I have read so far.

apps/server/src/services/dictionaryService.ts
```diff
@@ -211,3 +212,8 @@ export const updateSchema = async (id: string, schema: Schema, major: boolean):
 	// Save new dictionary version
 	return await DictionaryRepo.addDictionary(omit(updatedDictionary, '_id'));
 };
+
+export const downloadDictionaryByNameAndVersion = async (name: string, version: string) => {
+	const dict = await getOneByNameAndVersion(name, version);
+	return replaceReferences(dict);
```
joneubank, line 218: What does this function do different from `getOneByNameAndVersion`? Is it just the reference replacement? It's not actually providing a download or anything like that.

apps/server/src/controllers/dictionaryController.ts
```diff
@@ -133,19 +135,67 @@ export const getSchemaField = async (
...
+
+		for (const schema of dictionary.schemas || []) {
+		const fields = schema.fields || [];
+			const templateRow = fields.reduce((acc: { [x: string]: string }, field: { name: string | number }) => {
+				acc[field.name] = '';
+				return acc;
+			}, {});
+
+		const tsvParser = new Json2tsv({ delimiter: '\t', quote: '' });
+		const tsv = tsvParser.parse([templateRow]);
```
joneubank, line 179: Can we move this to a function like `createDataFileTemplate(schema: Schema, options?: {delimiter?: string})` with default delimiter `\t`?

I'd like to include this functionality in the `lectern-dictionary` package, so that we could reuse this schema generating functionality in browser in the future. So let's add this function into `dictionary/src/dataFileTemplates.ts` and then export it from `dictionary/src/index.ts`. We can import it from there into this dictionaryController.

joneubank (changes requested):
Updates from feedback look good, thanks for that.

I have a refactoring request to move the dataFileSchema generation into the `lectern-dictionary` package, can we add that before we merge this?

packages/dictionary/src/dataFileTemplates.ts
```diff
@@ -0,0 +1,46 @@
...
+): { fileName: string; content: string } => {
+	const delimiter = options?.delimiter ?? '\t';
+	const fields = schema.fields || [];
+
+	const templateRow = fields.reduce((acc: { [key: string]: string }, field: { name: string }) => {
+		acc[field.name] = '';
+		return acc;
+	}, {});
+
+	const tsvParser = new Json2tsv({ delimiter, quote: '' });
```
joneubank, line 39: Looks like you meant to use the `quote` property but didn't. Should we update this? Either remove the option or use it (probably remove it).

packages/dictionary/src/dataFileTemplates.ts
```diff
@@ -0,0 +1,46 @@
...
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+import { Json2tsv } from 'json2tsv';
+import { Schema } from './types';
+
+interface CreateDataFileTemplateOptions {
+	delimiter?: string;
+	quote?: string;
```
joneubank, line 24: Remove this option if we aren't using it.

joneubank (changes requested):

packages/dictionary/src/dataFileTemplates.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export type DataFileTemplateConfig = { delimiter: string; extension: string };
+
+const SeparatedValueFileConfigs = {
+	csv: { delimiter: ',', extension: 'csv' },
+	tsv: { delimiter: '\t', extension: 'tsv' },
+} as const satisfies Record<SeparatedValueFileType, DataFileTemplateConfig>;
+
+export type CreateDataFileTemplateOptions = { fileType: SeparatedValueFileType } | DataFileTemplateConfig;
+export type DataFileTemplate = { fileName: string; content: string };
```
joneubank, line 34: Exporting the return type so that any consumer can use this in their code.

packages/dictionary/src/dataFileTemplates.ts
```diff
@@ -0,0 +1,76 @@
...
+ * TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
+ * OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER
+ * IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ * ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { z } from 'zod';
+import { Schema } from './metaSchema/index';
+
+export const separatedValueFileTypeSchema = z.enum(['tsv', 'csv']);
```
joneubank, line 23: Defining enum with Zod since I use this is useful for validating fileType selector on server endpoints.

packages/dictionary/src/dataFileTemplates.ts
```diff
@@ -17,32 +17,60 @@
...
+	tsv: { delimiter: '\t', extension: 'tsv' },
+} as const satisfies Record<SeparatedValueFileType, DataFileTemplateConfig>;
 
-export const createDataFileTemplate = (
-	schema: Schema,
-	options?: CreateDataFileTemplateOptions,
-): { fileName: string; content: string } => {
-	const delimiter = options?.delimiter ?? '\t';
-	const fields = schema.fields || [];
+export type CreateDataFileTemplateOptions = { fileType: SeparatedValueFileType } | DataFileTemplateConfig;
```
Azher2Ali, line 33: Can we consider discriminated union with a type property or It might be clearer if made two separate types:

packages/dictionary/src/dataFileTemplates.ts
```diff
@@ -17,32 +17,60 @@
...
+ * Create the contents of a file that can be used to store data that will be validated with the provided Schema.
+ * The file will be a separated values file with a header row. The headers correspond to the fields in the schema.
+ * An empty row with placeholder delimiters for each field is also included. Also returned is the lectern standard
+ * filename for the file, which will use the schema name.
+ *
+ * By default this file will be a tab separated value file (`.tsv`). If an options object is provided, you can
+ * either specify the `fileType` as `tsv` or `csv`. Selecting `csv` will provide a comma separated value file
+ * (`.csv`) instead.
+ *
+ * If you do not specify a file type, the options also yout to directly specify the file extension and delimiter value
```
Azher2Ali, line 46: Typo fix: 'allow you' to instead of 'yout to'

packages/dictionary/src/dataFileTemplates.ts
```diff
@@ -17,32 +17,60 @@
...
+ * that is used.
+ *
+ * @param schema
+ * @param options - `{ fileType: 'csv' | 'tsv' }` OR `{ delimiter: string; extension: string; } `
+ * - default if not provided will use 'tsv'
+ * @returns
+ */
+export function createDataFileTemplate(schema: Schema, options?: CreateDataFileTemplateOptions): DataFileTemplate {
+	const config = !options
```
Azher2Ali, line 55: Could be cleaner:
```
const config = options
    ? 'fileType' in options
        ? SeparatedValueFileConfigs[options.fileType]
        : options
    : SeparatedValueFileConfigs.tsv;
```

packages/dictionary/src/dataFileTemplates.ts
joneubank: Would be fine, I didn't do this cause I had no use for reusing that extra type. Can easily be refactored that way in the future if needed.

joneubank (approved):
Good work team. In a good spot to make this available in the dictionary package and server.

We want to followup with proper unit tests for the template generation: http://github.com/overture-stack/lectern/issues/254

PR #290, opened 2025-06-09
Feature/interaction panel

SaqAsh:
## Issues
- #(issue number)
- #(issue number)

## Readiness Checklist

- [ ] **Self Review**
- [ ] **Formatting**
- [ ] **Local Testing**
- [ ] **Updated Tests**
- [ ] Documentation


packages/ui/package.json
SaqAsh, line 1: Needed the lectern client based types, thus added lectern client

packages/ui/src/viewer-table/InteractionPanel/InteractionPanel.tsx
```diff
@@ -0,0 +1,106 @@
...
+				<DownloadTemplatesButton
+					fileType="tsv"
+					version={currentDictionary.version}
+					name={currentDictionary.name}
+					lecternUrl={currDictionary.lecternUrl}
+					disabled={disabled}
+				/>
+			</div>
+		</div>
+	);
```
ciaranschutte, line 103: this is looking super nice

ciaranschutte (changes requested):
couple of small changes to start with.

I'll take a look if there's a cleaner way to do the disabled state being passed in as props everywhere. a context provider might make sense, but adds some complexity. 

There's also maybe a slightly cleaner way to compose css with emotion that I will verify.
Because emotion serializes to classNames, you basically create `css..` fragments and conditionally apply them using https://www.npmjs.com/package/clsx

packages/ui/src/common/Dropdown/Dropdown.tsx
```diff
@@ -23,13 +23,15 @@
...
 import DropDownItem from './DropdownItem';
 
-const dropdownButtonStyle = (theme: Theme, width?: string) => css`
+const dropdownButtonStyle = (theme: Theme, width?: string, disabled?: boolean) => css`
 	display: flex;
 	flex-wrap: nowrap;
+	white-space: nowrap;
 	align-items: center;
 	justify-content: center;
 	gap: 11px;
```
ciaranschutte, line 37: bit of a cleaner way because there are additional styles based on a state "disabled",
we can move the those disabled styles into their own declaration.
then conditional apply them if needed in the main style

```tsx

const disabledButtonStyle = css`
	cursor: not-allowed;
	opacity: 0.7;
`;

const dropdownButtonStyle = ({ theme, width, disabled }) => css`
	display: flex;
	flex-wrap: nowrap;
	white-space: nowrap;
	align-items: center;
	justify-content: center;
	gap: 11px;
	width: ${width || 'auto'};
	min-width: fit-content;
	padding: 8px 16px;
	background-color: #f7f7f7;
	color: ${theme.colors.accent_dark};
	border: 1px solid #beb2b294;
	border-radius: 9px;
	height: 42px;
	box-sizing: border-box;
	cursor: pointer;
	transition: all 0.2s ease;
	z-index: 1000;
	opacity: 1;

	${disabled && disabledButtonStyle}
`;
```

if you change the params to object destructuring params like:
```tsx
const dropdownButtonStyle = ({ theme, width, disabled }) => css`
```
then you don't have to worry about passing in "undefined" eg. for width, and you don't have to worry about order

usage:
```tsx
<div css={dropdownButtonStyle({ theme, disabled })} onClick={handleToggle}>
```

ciaranschutte (changes requested):
looking good. requesting changes because I'll have to re-approve anyway ;)

ciaranschutte (approved):

Issue #315, opened 2025-07-16
5E Rendering Allowed Values


MitchellShiell:
The basic pattern would primarily require a `must be` statement followed by the allowed values as rendered in the table. 

```
IF [field_referenced] [condition]
    THEN [conditional_field] must be [allowed values]
    ELSE must be [allowed values] (if applicable)
```

> [!IMPORTANT]
> One exception will be how we handle codelists, i've detailed this with a mock up provided below

## Basic Pattern Examples

**CodeList (multiple options):**
```
THEN image_pipeline must be one of: pipeline_A, pipeline_B, pipeline_C
```

**CodeList (single option):**
```
THEN image_pipeline must be: pipeline_A
```

> [!NOTE]
> The `codelist` allowed values verbiage in the table may need a minor update to conform with this

**Required field:**
```
THEN image_pipeline must be provided
```

**Range (numeric):**
```
THEN age must be between 18 and 65
THEN count must be greater than 0
```

**Regex pattern:**
```
THEN identifier must match pattern: ^NCIT:C\d+$
```

**Boolean:**
```
THEN is_active must be true
THEN is_active must be false
```

**Empty restriction:**

We may not have accounted for this in the table as it only really is relevant as an allowed value following a condition

```
THEN comments must be empty
```

The "must be" pattern works grammatically with all these cases. This creates a consistent, readable format that users can quickly scan and understand.

## Rendering Codelist 

Codelists will need to be rendered differently for the modal. For long code lists render as follows:

<img width="2968" height="2132" alt="Image" src="https://github.com/user-attachments/assets/cb0df51b-f77c-4559-98f9-d7497af860e8" />

https://www.figma.com/design/LILnnxiWOSiDF7k7IhBYmj/Overture_DataDictionary?node-id=3609-1651&t=ht5WxX3Ztzc3JAuv-4

and for shorter codelists:

<img width="2968" height="2132" alt="Image" src="https://github.com/user-attachments/assets/8854d9a1-3154-411f-8221-49c9a09b1081" />

https://www.figma.com/design/LILnnxiWOSiDF7k7IhBYmj/Overture_DataDictionary?node-id=3640-735&t=ht5WxX3Ztzc3JAuv-4

SaqAsh:
Currently there is no implementation for codeLists with count restrictions, does not exist on the type. 
Not sure on what the `Boolean` field is, would need clarification since Boolean does not exist on the type.

PR #329, opened 2025-07-22
#315: Render allowed values inside of modal

SaqAsh:
## Summary
Implemented RenderAllowedValues component and ListItem styling component to display schema field restrictions.

## Issues
- #315 

### UI
- Added a new component `ListItem` which creates styled pill badges for displaying individual code list values.
  - Description of `ListItem`: A reusable styled component built on top of `Pill.tsx` that is used to render individual items as small gray pills. Going to be used inside of the conditional logic modal 
  - Added a new component `RenderAllowedValues` which specifically renders the allowed values for a given schema field

## Readiness Checklist

- [X] **Self Review**
- [X] **Formatting**
- [ ] **Local Testing**
- [ ] **Updated Tests**
- [X] Documentation


packages/ui/src/common/ListItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+	fontFamily: "'DM Mono', monospace",
+	borderRadius: '8px',
+	width: 'fit-content',
+	maxWidth: 'none',
+	backgroundColor: '#ECECEC',
+	border: '0.5px solid black',
+	display: 'flex',
+	alignItems: 'center',
+	justifyContent: 'center',
+});
```
SaqAsh, line 45: Although it looks like theme is not being used here ( it is not, it will be used when we just add all the appropriate fonts into the theming

packages/ui/src/common/ListItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+const getListItemStyles = (theme: Theme): CSSProperties => ({
+	fontSize: '13px',
+	fontWeight: 'normal',
+	lineHeight: '1',
+	fontFamily: "'DM Mono', monospace",
+	borderRadius: '8px',
+	width: 'fit-content',
+	maxWidth: 'none',
```
ciaranschutte, line 39: pretty sure this is the default.

packages/ui/src/common/ListItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useThemeContext } from '../theme/ThemeContext';
+import Pill from './Pill';
+
+export interface ListItemProps {
+	children: ReactNode;
+}
+
+const getListItemStyles = (theme: Theme): CSSProperties => ({
+	fontSize: '13px',
+	fontWeight: 'normal',
```
ciaranschutte, line 34: this should be default already

packages/ui/src/common/ListItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+	lineHeight: '1',
```
ciaranschutte, line 35: unless we're explicitly setting this for something, let's remove it. I've seen in a bunch of other PRs across the projects we tend to just copy pasta in Figma snippets, and it's overcomplicating the actual code. Things like font-family and line-height should be inherited for the most part

packages/ui/src/common/ListItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+	const theme: Theme = useThemeContext();
+
+	return (
+		<Pill size="small" style={getListItemStyles(theme)}>
+			{children}
+		</Pill>
+	);
+};
+
+export default ListItem;
```
ciaranschutte, line 57: looking nice!

packages/ui/src/viewer-table/ConditionalLogicModal/RenderAllowedValues.tsx
```diff
@@ -0,0 +1,145 @@
...
+ *  IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/** @jsxImportSource @emotion/react */
+
+import { SchemaField, SchemaFieldRestrictions } from '@overture-stack/lectern-dictionary';
+import React, { Fragment } from 'react';
+
+import { css } from '@emotion/react';
```
ciaranschutte, line 25: pop this up one new line with the external libs please, keep linebreak between 3rd party libs and internal components

ciaranschutte (changes requested):
looking really good! a couple of minor tweaks. I think you also need to add the font.
When you set a `font-family` style in CSS it's referencing fonts it has access to, it isn't downloading anything. It will always default to something on your system.

So in this case if you want to use DM Mono, you'll have to a download to it.
usually something like this in the head of the document
```
 <link
        href="https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,100;0,300;0,400;0,700;0,900;1,100;1,300;1,400;1,700;1,900&family=Work+Sans:ital,wght@0,100..900;1,100..900&display=swap"
        rel="stylesheet"
      ></link>
 ```
 
 we use Lato a bunch in this package too, and it doesn't seem to be imported yet either

File: packages/ui/.storybook/preview.tsx [hunk 7]
```diff
@@ -1,9 +1,24 @@
...
+
+			const latoLink = document.createElement('link');
+			latoLink.href =
+				'https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,100;0,300;0,400;0,700;0,900;1,100;1,300;1,400;1,700;1,900&display=swap';
+			latoLink.rel = 'stylesheet';
+			document.head.appendChild(latoLink);
+
+			return <Story />;
+		},
+	],
```
SaqAsh, line 21: don't need to add if we add `preview-head.html`

ciaranschutte (approved):

PR #333, opened 2025-08-01
#260 - dictionary controller

SaqAsh:
## Summary

Added a dictionary data controller with React context for managing dictionary data from multiple sources with state management and filtering.

## Issues
- #260 
### UI
- **DictionaryDataContext**: Created a centralized React context for dictionary data management
  - Supports three data source types: static dictionaries, remote Lectern API, and hosted URLs
  - Provides loading states, error handling, and dictionary version management
  - Includes filtering capabilities for required vs all fields
- **Data Sources**: Implemented modular data fetching layer
  - `lectern.ts`: Handles remote dictionary fetching from Lectern API with version support
  - `hosted.ts`: Manages hosted dictionary validation and fetching with Zod schema validation
- **Context Hook**: Added `useDictionaryDataContext` hook with proper error boundaries
- **ErrorModal**: Enhanced error handling component in common directory
  - Displays error messages in a structured list format with icons
  
## Storybook 
- Created a `dictionaryDecorator` that provides mock dictionary data through the `DictionaryDataProvider` context, allowing Storybook stories to simulate different states without duplication. 
- Avoid having to individually wrapping stories with the context

## Testing

- This was testing using the `make stage-dev` command on [prelude](https://github.com/overture-stack/prelude)
## Readiness Checklist

- [X] **Self Review**
- [X] **Formatting**
- [X] **Local Testing**
- [ ] **Updated Tests**
- [X] Documentation

### DIFF
--- a/packages/ui/src/dictionary-controller/DictionaryDataContext.tsx
+++ b/packages/ui/src/dictionary-controller/DictionaryDataContext.tsx
@@ -20,7 +20,7 @@
 
 import type { DictionaryServerRecord } from '@overture-stack/lectern-client/dist/rest';
 import type { Dictionary } from '@overture-stack/lectern-dictionary';
-import React, { createContext, ReactNode, useContext, useEffect, useState } from 'react';
+import { createContext, ReactNode, useContext, useEffect, useState } from 'react';
 
 import { fetchAndValidateHostedDictionaries, fetchRemoteDictionary } from './sources';
 
@@ -41,6 +41,7 @@ export type DictionaryStateContextType = {
 	filters: FilterOptions[];
 	setCurrentDictionaryIndex: (index: number) => void;
 	setFilters: (filters: FilterOptions[]) => void;
+	selectedDictionary?: DictionaryServerUnion;
 };
 
 export type StaticDictionaryProviderProps = {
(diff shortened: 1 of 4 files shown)


packages/ui/src/common/ErrorModal.tsx
```diff
@@ -0,0 +1,159 @@
...
+		boxShadow: '0px 4px 4px 0px #0000001A',
+		display: 'flex',
+		flexDirection: 'column',
+		backgroundColor: theme.colors.error_modal_bg,
+	},
+	overlay: {
+		backgroundColor: 'transparent',
+		zIndex: 1000,
+	},
+});
```
ciaranschutte, line 55: please add a small comment that these are for the modal vs our normal Emotion css style. this has tripped me up on another project wondering why we're suddenly nesting objects and using strings

packages/ui/src/dictionary-controller/DictionaryDataContext.tsx
```diff
@@ -0,0 +1,165 @@
...
+		name: 'dictionaryName' in props ? props.dictionaryName : undefined,
+		loading,
+		errors,
+		currentDictionaryIndex,
+		filters,
+		setCurrentDictionaryIndex,
+		setFilters,
+	};
+
+	return <DictionaryDataContext.Provider value={value}>{props.children}</DictionaryDataContext.Provider>;
```
ciaranschutte, line 164: this is one of those times where I'd choose simplicity over a more complicated piece of conditional code.

The pattern of "checking if we have <x> prop, then all functionality based on that conditional" is a good indicator these can be seperate components. 

How does it look if we separate data loading and "dictionary state".
like....

```tsx
<DictionaryStaticDataProvider>
  <DictionaryController/>
</DictionaryStaticDataProvider>
// different usage
<DictionaryUrlDataProvider>
  <DictionaryController/>
</DictionaryUrlDataProvider>
// .. and so on
```

This way data fetching along with states (loading, errors etc) is constrained to one component.
This component doesn't need to check for valid props, it's up to the consumer to use the correct component.
Then you can pass that state down to `DictionaryController` if its valid state using a context provider...
```tsx
...
const [filters, setFilters]  =  useState(filters);
const {dictionaryData, isLoading, errors} = useDictionaryData();

const data = filterData(filters, dictionaryData)
...
```

All variable naming etc is just demonstrative of the bigger idea to move data provding into individual components :)
We can run through on Monday

packages/ui/src/dictionary-controller/DictionaryDataContext.tsx
```diff
@@ -0,0 +1,165 @@
...
+					switch (true) {
+						case err instanceof Error:
+							message = err.message;
+							break;
+						case typeof err === 'string':
+							message = err;
+							break;
+						default:
+							message = 'Something went wrong';
+					}
```
ciaranschutte, line 141: `switch(true)` is an antipattern
the `case` is meant to be the conditional you're checking in the switch(..) call
using it this way is a more obtuse if/else... so you can just use an if / else ;)

ciaranschutte (changes requested):
Some really good work! nice abstracting the data fetching and the storybook improvements are great. comments and types very much appreciated.

A few bits of re-arranging to do that we can step through this week

PR #343, opened 2025-08-29
release lectern-ui v0.0.1-beta.0

justincorrigible:
adds all the changes from `main`, as well as bumps versions for all sibling packages.


scripts/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "lectern-scripts",
-	"version": "2.0.0-beta.5",
+	"version": "2.0.0-beta.6",
```
joneubank, line 3: This `scripts` package isn't meant to be imported anywhere... it could maybe have the version removed like the root dir?

joneubank (approved):

scripts/package.json
justincorrigible: amending here

Issue #416, opened 2026-03-02
Table View Styling Updates


MitchellShiell:
## Tool bar updates

- Update button heights to 30px
- Stroke weight to 1px
- Fonts to 16px regular
- Reduce weight of icons (svgs provided below)
- Move Diagram view button next to table of contents dropdown

<img width="2886" height="120" alt="Image" src="https://github.com/user-attachments/assets/4f8fe2d8-3176-488d-84f4-c8dfc51a609f" />

### Icons

Version icon:

```
<svg width="13" height="13" viewBox="0 0 13 13" fill="none" xmlns="http://www.w3.org/2000/svg">
<path d="M0.5 6.5C0.5 7.68669 0.851894 8.84672 1.51118 9.83342C2.17047 10.8201 3.10754 11.5891 4.2039 12.0433C5.30026 12.4974 6.50665 12.6162 7.67054 12.3847C8.83443 12.1532 9.90352 11.5818 10.7426 10.7426C11.5818 9.90352 12.1532 8.83443 12.3847 7.67054C12.6162 6.50665 12.4974 5.30026 12.0433 4.2039C11.5891 3.10754 10.8201 2.17047 9.83342 1.51118C8.84672 0.851894 7.68669 0.5 6.5 0.5C4.82263 0.50631 3.21265 1.16082 2.00667 2.32667L0.5 3.83333M0.5 3.83333V0.5M0.5 3.83333H3.83333M6.5 3.16667V6.5L9.16667 7.83333" stroke="#003055" stroke-linecap="round" stroke-linejoin="round"/>
</svg>
```

diagram view (eye):

```
<svg width="18" height="13" viewBox="0 0 18 13" fill="none" xmlns="http://www.w3.org/2000/svg">
[...]

PR #448, opened 2026-07-30
Overture docs update

MitchellShiell:
Documentation for the [Overture docs site](https://github.com/overture-stack/docs). No source code changes; the only non-`docs/` change is adding `.DS_Store` to `.gitignore`.

`docs/04-lecternViewer.md` becomes a three-page section, `docs/04-viewer/`, written against `packages/ui` as it stands on `main`.

| File | Sidebar label | URL |
| --- | --- | --- |
| `docs/04-viewer/index.md` | Overview | `/develop/Lectern/viewer/` |
| `docs/04-viewer/01-setup.md` | Viewer Setup | `/develop/Lectern/viewer/setup` |
| `docs/04-viewer/02-reference.md` | Viewer Reference | `/develop/Lectern/viewer/reference` |

## Other changes

- `docs/01-overview.md`: Lyric links repointed; repository tree corrected (`common` removed, `ui` added); components table de-indented so it renders, and the Dictionary/Validation npm badges and paths fixed, both having pointed at the client package.
- `docs/02-Setup.md`: documents the optional `MONGO_URL`, which takes precedence over `MONGO_HOST`/`MONGO_PORT`/`MONGO_DB`.
- READMEs: links into the deleted `develop` branch repointed at `main`, and `docs/important-concepts.md#testresult` repointed at `pendingDocs/glossary.md#testresult`.
- "Need Help?" blocks repointed to `/community/support`; `docs/assets/submission-system.svg` updated.

## Rebased onto `main`

Current with `main`, no conflicts. 43 commits absorbed, with one conflict in `README.md` where both sides had repointed the same stale link; this branch's `/develop/Lectern/overview` was kept over `main`'s older `docs/core-software/` scheme. `main`'s reformat of `docs/03-dictionaryReference.md` and its `displayName` additions are untouched. The viewer pages were re-checked against the rebased `packages/ui`: entry points, `DictionaryTableProps`, `CustomColumnConfig` and `MetaValueRenderer` all match.

## Known issues, not fixed here

- Screenshots predate `#416: Table View Styling Updates`. Recapturable from Storybook, but not needed for this pass.
- `packages/ui/src/index.ts` aliases `SchemaTable` and `Toolbar` to the same `default` export as `DictionaryTable`, so all three resolve to `DictionaryViewerPage`. Also in the published `1.0.0` types. The Reference page routes readers to the `/dictionary-table` subpath instead, and notes the caveat.
- `Columns/DataType.tsx` renders `Array` in place of the value type, so array-of-string and array-of-integer look identical.
- `packages/ui/README.md` and `packages/ui/docs/README.md` still advertise the removed `LecternThemeProvider`/`LecternTheme` names.

## Broken code fence fixed

`docs/03-dictionaryReference.md` had an unclosed four-backtick fence in the "Display Name Example" block, which arrived with the `displayName` work on `main`. It swallowed the `## Field Restrictions` heading and everything after it into a code block, so the `#field-restrictions` anchor stopped existing and the Docusaurus build failed on two broken anchors: the field-properties table on this page, and the docs site's `/use/administration/building-dictionaries` page.

Originally flagged here as out of scope, then fixed once it broke the site build, since this branch already owns the file.

## CI

The Jenkins failure is unrelated to this branch, which changes no build inputs. The shared pipeline runs `npx --yes pnpm install`, which resolves to pnpm 11, requiring Node >= 22.13; the agent runs Node 18.16.1. `main` fails identically.


docs/04-viewer/01-setup.md
```diff
@@ -0,0 +1,154 @@
+# Viewer Setup
+
+This guide covers embedding the Lectern Viewer in a React application: installing the package, pointing it at a dictionary source, and configuring filters, columns, and theming. For what the resulting views show a reader, see the [Viewer overview](./index.md); for the complete export listing, see the [Reference](./02-reference.md).
+
+## Prerequisites
+
+- **React 19.** The package declares `react` and `react-dom` `^19.1.0`.
+- **A dictionary source.** Either a running [Lectern server](../02-Setup.md), a dictionary JSON file served over HTTP, or dictionary objects already in your application.
+- **Emotion.** Components style themselves with `@emotion/react`, which ships as a dependency. Applications using a different CSS-in-JS library can still mount the components, but theming goes through Emotion.
```
joneubank: This is not a prerequisite, its just a dependency that the developer is going to have to work with. Not sure it should be here or in a later section on "theming"

docs/04-viewer/01-setup.md
```diff
@@ -0,0 +1,154 @@
...
+   pnpm --filter @overture-stack/lectern-ui storybook
+   ```
+
+Storybook runs on port `6006` by default: [http://localhost:6006/](http://localhost:6006/).
+
+Stories render under a `themeDecorator`, and a theme selector in the Storybook toolbar applies an alternate theme to every story that uses it — useful for checking that a component reads the theme rather than hard-coding values.
+
+![Storybook theme selector in the toolbar](../assets/global-theme-selector.png)
+
+For the full walkthrough — adding themes to the selector, wiring the decorator, and editing stories — see the [Lectern UI developer docs](https://github.com/overture-stack/lectern/blob/main/packages/ui/docs/README.md) in the repository.
```
joneubank: Place the storybook section after the intro or install sections since it is the reference documentation?

We should consider running the storybook on a publicly hosted URL so that devs can see whats available before going through the install.

docs/04-viewer/02-reference.md
joneubank, line 1: Impressive documentation. IMO its the storybook that is responsible for documenting the components available but its nice having this reference as well. As per usual, my concern is that it will be a challenge to maintain both to same standard and that leads to reduced trust in the library.

docs/04-viewer/02-reference.md
```diff
@@ -0,0 +1,179 @@
...
+
+### `DictionaryTableStateProvider`
+
+Holds view state — the selected version, the required-fields filter, and active metadata filter selections — and must wrap the view inside whichever data provider is used. Takes only `children`. Its prop type is exported as `DictionaryTableStateProviderProps`.
+
+Changing the selected version clears any active metadata filter selections, since a new version may not carry the same metadata values.
+
+## Hooks
+
+Both hooks throw when called outside their provider.
```
joneubank: This should be like an Important! callout. It should probably live, in duplicate, in each of the hook sections and not here.

docs/04-viewer/02-reference.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Theming
+
+| Export | Type | Description |
+| --- | --- | --- |
+| `ThemeProvider` | Component | Supplies a theme. Takes an optional `theme` prop, deep-merged into the active theme, so partial overrides are valid. |
+| `defaultTheme` | `Theme` | The theme used when no provider is present. |
+| `useThemeContext` | `(overrides?: PartialTheme) => Theme` | Reads the active theme, optionally merging per-call overrides. |
+| `Theme` | Type | The full theme: `colors`, `typography`, `dimensions`, `shadow`, and `icons`. |
+| `PartialTheme` | Type | A recursively partial `Theme`, for passing to `ThemeProvider`. |
```
joneubank, line 169: Do we document what are the properties available in the theme and what htey control? don't NEED to add that to this PR, its a very lofty goal, but would be sweet.

docs/01-overview.md
```diff
@@ -12,7 +12,7 @@ Data dictionaries are organized collections of schemas that define the structure
 
 ## System Architecture
 
-Lectern operates as a central Dictionary Schema repository, providing dictionary management and validation services through its RESTful API. The service maintains schemas in a database (mongoDb), tracking versions and relationships between different schema elements. In the Overture platform Lectern's schemas are consumed by [Lyric](https://docs.overture.bio/docs/under-development/lyric/), which stores and uses them to validate incoming tabular data submissions.
+Lectern operates as a central Dictionary Schema repository, providing dictionary management and validation services through its RESTful API. The service maintains schemas in a database (mongoDb), tracking versions and relationships between different schema elements. In the Overture platform Lectern's schemas are consumed by [Lyric](https://docs.overture.bio/develop/Lyric/overview), which stores and uses them to validate incoming tabular data submissions.
```
joneubank:
```suggestion
In the Overture platform, Lectern operates as a central Dictionary Schema repository, providing dictionary management and validation services through its RESTful API. The service maintains schemas in a database (mongoDb), tracking versions and relationships between different schema elements. [Lyric](https://docs.overture.bio/develop/Lyric/overview) references schemas hosted in Lectern and uses them to validate incoming tabular data submissions.
```

joneubank (changes requested):

docs/04-viewer/01-setup.md
MitchellShiell: Moved into the theming section.

docs/04-viewer/01-setup.md
MitchellShiell: Moved to just after the intro. Hosted Storybook will likely be needed, its relevant and can be linked into the dictionary viewer paper i am currently writing too

docs/04-viewer/02-reference.md
MitchellShiell: Done. One callout per hook section, each naming its own provider.

docs/04-viewer/02-reference.md
MitchellShiell: Added a table with the five theme groups, what each controls, w/ link to source

docs/04-viewer/02-reference.md
MitchellShiell: agreed maybe once we have the storybook hosted and linked then we can revisit these docs and how much of it we want to keep/maintain

docs/04-viewer/02-reference.md
```diff
@@ -158,6 +168,18 @@ The loading skeletons — `LoadingSpinnerPage`, `HeaderSkeleton`, `ToolbarSkelet
 | `Theme` | Type | The full theme: `colors`, `typography`, `dimensions`, `shadow`, and `icons`. |
 | `PartialTheme` | Type | A recursively partial `Theme`, for passing to `ThemeProvider`. |
 
+### Theme structure
+
+`Theme` is inferred from `defaultTheme`, so the groups below are the complete set of what a `PartialTheme` can override. Each links to its definition, which is the authoritative list of tokens and their values.
```
joneubank:
```suggestion
`Theme` is inferred from [`defaultTheme`](https://github.com/overture-stack/lectern/blob/main/packages/ui/src/theme/defaultTheme.ts), so the groups below are the complete set of what a `PartialTheme` can override. Each links to its definition, which is the authoritative list of tokens and their values.
```

joneubank (dismissed):

joneubank (approved):

